Issue #32, opened 2014-12-12
Store extension along with id


jnicklas:
It might be useful to know the file extension the file had when it was uploaded. My current thinking is that we should carry this around as part of the `id` in attachments. So an id might be `a1bca.jpg` or maybe `a1bca/jpg` or something. This would make it interact in a nicer way with multiple file uploads in the future.

This would mean we would have a different notion of what an `id` is in the attachment part of Refile, versus the backend part. This is possibly a little weird, but I think it outweighs the cons.

There remains a question whether this should be:

a) done at all
b) be opt-in
c) always be done

I'm leaning towards (c).

PR #70, opened 2014-12-26
Store additional metadata, filename, size, etc...

jnicklas:
See also #32 and #8

This is a work-in-progress PR that I'm sharing early for feedback. I'm trying to get Refile to store file metadata, namely filename, size and content type in the DB alongside the id. This is particularly tricky because it **must** work with all of Refile's existing features:
- [x] Direct and presigned uploads
- [x] Uploads from URLs
- [x] Form redisplays
- [ ] Extension/content_type validations
- [x] ActiveRecord integration
- [x] Removing attached files


asiniy:
So, when this update will released in the gem? And, how can I now the extension of the file?

PR #82, opened 2014-12-29
Extract filename and content type into Refile::Uploadable

janko-m:
Instead of extracting filename and content_type through the Refile constant, it feels much better to take an object oriented approach. Refile::File now wraps the uploadable from the outside world, and doesn't know anything about backends or IDs.

What I originally intended with this is that we can now, instead of handling content types and extensions in `Refile::Attacher`, just let users register their validators (and pass them though `:type` parameter), which would yield a `Refile::File` and then users can write their own validation (the same way that they register processors). Of course, Refile would still provide an `:image` validator. This way `Refile::File` would abstract the "uploadable".

I realized only before opening this PR that I broke the `#copy_to` in the S3 backend. Do you think this approach has potential to justify trying to fix the S3 in another way? And also justify `Refile::Backend#upload` now returning an ID instead of a `Refile::File`? Probably not :P

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,6 +28,9 @@ Style/StringLiterals:
 Style/StringLiteralsInInterpolation:
   EnforcedStyle: double_quotes
 
+Style/GuardClause:
+  Enabled: false
+
 Style/AndOr:
   Enabled: false
 
--- a/Rakefile
+++ b/Rakefile
@@ -12,7 +12,9 @@ end
 
 RSpec::Core::RakeTask.new(:spec)
 
-RuboCop::RakeTask.new
+RuboCop::RakeTask.new do |task|
+  task.options = ["--display-cop-names"]
+end
 
 task default: [:spec, :rubocop]
 
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -2,7 +2,6 @@ require "uri"
 require "fileutils"
 require "tempfile"
 require "logger"
-require "mime/types"
 
 module Refile
   # @api private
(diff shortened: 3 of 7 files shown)


jnicklas:
This is interesting. It's a very large change, so difficult to review properly.

This helps a lot in solving the currently biggest design problem in Refile which is the Attachor. The code in it is pretty bad, and it's gotten even worse since we're juggling with metadata and validations. Improving the Attachor is high on my list.

That being said, I specifically chose not to wrap uploadables in a class. [I did this](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/sanitized_file.rb) in CarrierWave and the result is incomprehensible and bad. There's a lot of juggling in CarrierWave when files are wrapped in layers and layers of wrappers. It's pretty hard to understand what's going on. My idea was to go for a more functional approach of data flowing through, rather than OO, but this turns out a little weird in Ruby.

IMO, As implemented this PR makes the Backend API too low level and finicky. One of my design goals with Refile was that backends should be useful and nice to work with in of themselves. Having real File objects, instead of passing around ids feels much cleaner and Ruby-esque to me.

However, the idea of having `Refile::File` play this dual role is a bit dangerous too. This ends up in the same kind of hell that `SanitzedFile` is in. Should the metadata form of files also be `Refile::File`? It becomes a wrapper for everything. Also it doesn't make any sense, since for example an uploadable file can be asked for its content_type, whereas a file stored in a Refile backend explicitly does *not* know what its content type is, by design.

Should we add two different wrappers then, maybe `Refile::File` and `Refile::Uploadable` or something? Maybe. This seems like the least bad option to me, but it brings us dangerously close to the kind of confusion that's inherent in CarrierWave.

Also, consider this: an `ActionDispatch::UploadedFile` already exactly conforms to the interface we want. We can ask it for content_type, size, filename, read from it, copy_stream it, close it. That's all we need, really. Remember also that `ActionDispatch::UploadedFile` already *is* a wrapper around the raw Rack file upload data. In the most common case of what we're receiving, we wrap a wrapper in another wrapper which is identical to the original wrapper. Blurgh.

So the current code isn't super nice, but is wrapping it in OO better? Maybe. But we should consider it carefully before we run down that route. Just because there is less code doesn't necessarily mean it's a better design.

So this isn't really a definite yes or no to this PR. It's a maybe, I guess.

janko-m:
Thank you, I really appreciate the detailed review.

I agree, the dual role could bring back the `CarrierWave::SantizedFile` hell, so it's definitely a no-go.

However, I really like the idea of a separate `Refile::Uploadable` class. And I can argue that it wouldn't be anything like `CarrierWave::SanitizedFile`. The main reason is that you narrowed the `uploadable` interface upfront to specified 4 methods, which means the implementation of `Refile::Uploadable` won't be able to grow after the initial version.

I also see as a great benefit that `Refile::Uploadable` would *force* the interface to the specificed 4 methods, so then backends will always throw errors if something is wrong, and currently they may throw or not throw errors depending on which file they get (they might accidentally be using methods which they shouldn't).

I remade the PR with `Refile::Uploadable`. The code changed a bit since 2 weeks ago, and I don't think it will help with simplifying `Refile::Attacher` (which I would really like), but it feels more OO.

jnicklas:
As a counter-point to this, we could have simply changed the various `extract_` methods into one `extract_metadata` which returns a Hash, and then assign this hash to `metadata`. The code would have been even cleaner in `Attacher` and we wouldn't even need the `Uploadable` wrapper.

janko-m:
Well, I don't think `extract_metadata` would make things better, because you would still have to split it into multiple private methods, which in my opinion shouldn't be on the `Refile` module.

That being said, since this you're probably thinking in the context of refactoring the `Attacher`, which I don't have knowledge of, I'm assuming I'm not seeing the bigger picture, so you can close this if you have a different plan.

jnicklas:
I've thought about this some more. What do you think about adding a class called `Refile::Metadata` whose sole purpose is extracting the metadata from an uploadable object.

It would be an internal class (`@api private`), and it would look pretty similar to your `Uploadable` class, but it would not wrap the IO methods, and it would not be passed to the backends, only used in the attacher to extract metadata from the uploadable object.

It would have a `to_h` method which would return the full metadata hash (minus id).

This way we don't end up with tons of wrapping, and we have a nice place to isolate all the metadata extraction. The validation is gone from the `Refile` singleton class now anyway. What do you think?

janko-m:
I got a bit out of sync with Refile the past few months. Since it's a small change anyway, I think it's best if I leave it to you :)

PR #84, opened 2014-12-29
CustomLogger now properly uses Refile.logger

### LINKED ISSUE #83: CustomLogger uses env["rack.errors"] instead of Refile.logger
This messes up Rspec output and log data is not available in rails log file.

krists:
Fixes #83

### DIFF
--- a/lib/refile/custom_logger.rb
+++ b/lib/refile/custom_logger.rb
@@ -3,14 +3,13 @@ module Refile
   class CustomLogger
     LOG_FORMAT = %(%s: [%s] %s "%s%s" %d %0.1fms\n)
 
-    def initialize(app, prefix = nil, logger = nil)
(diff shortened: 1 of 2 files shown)


lib/refile/custom_logger.rb
```diff
@@ -31,5 +29,14 @@ def log(env, status, began_at)
         (now - began_at) * 1000
       )
     end
+
+    def logger
+      @logger ||= @logger_proc.call
```
jnicklas: Was a bit fast on reviewing this. Lazily initializing like this is not thread safe. Probably not a huge deal, but it could lead to garbled logs in some rare cases.

krists: Worst case scenario would be that multiple threads would run `proc { Refile.logger }` multiple times, but `Refile.logger` is still the same object. I don't see any way how this would harm anything.

PR #86, opened 2014-12-29
Cache gem dependencies on Travis

janko-m:
With their new infrastructure also came this ability for open source projects, and bundling itself takes more than 1 minute.


jnicklas:
Nice :) Awesome!

PR #88, opened 2015-01-01
Initial specs for the Refile Image Processor

bry:
Initial Ruby specs for the Refile Image Processor only (before Image Processor refactor)
- Added mini_magick gem as development dependency
- Added color spec helper

Rubocop on `image_processing_spec.rb`, no offenses detected. All specs pass.
screen shot 2015-01-01 at 2 30 12 am


asiniy:
Nice dog!

bry:
@jnicklas I'm glad to contribute.

>Do you own the rights to the example image? If not, what usage rights have been granted by the author? We should use something which is public domain preferably.

I don't own the rights to the example image and I don't have the usage rights granted by the original author. 

I did [Google Image Search](https://www.google.com/search?as_st=y&tbm=isch&hl=en&as_q=cute+corgi+snow+paw&as_epq=&as_oq=&as_eq=&cr=&as_sitesearch=&safe=images&tbs=sur:fmc#imgdii=_) a replica of the image (2nd image down on the left) with different dimensions via [Advanced Image Search](https://www.google.com/advanced_image_search?as_st=y&hl=en&bav=on.2,or.r_cp.r_qf.&bvm=bv.83339334,d.cGU&biw=1157&bih=1037&dpr=1&q=cute+corgi+snow+paw&tbs=sur:fmc&tbm=isch) that is public domain and free to use, share or modify, even commercially. I'll use that one.

>A nice way of writing specs for image manipulation could be to do image comparison via perceptual hashes. Maybe that's a silly idea though, what do you think?

Yes, I think perceptual hashes are a nice and creative way leading to more solid, accurate specs.

>It would be nice to have tests for the behaviour of sending format to call.

I'll forge more tests around the behavior of sending `format` to `call`. Looks like there may be room for misuse of `call` when using the `format` parameter.

>Maybe we don't need to test every option in call, one or two should be enough.

Agree :smile:

bry:
@asiniy I agree :100:

jnicklas:
I never managed to get this merged, I tried a couple of times, but always ran into problems running this locally. In the end I ended up writing new tests from scratch. We now have tests for all the image processing operations in the refile/refile-mini_magick repo.

Thank you for putting in so much effort into this, and I'm sorry we didn't end up using it.

bry:
:+1:

PR #93, opened 2015-01-06
JRuby compat

hashrocketeer:
JRuby 9.0.0.0 is just around the corner, and targets the same featureset as Ruby 2.2. Only a couple small modifications were needed for compatibility.

### DIFF
--- a/lib/refile/backend/file_system.rb
+++ b/lib/refile/backend/file_system.rb
@@ -34,8 +34,12 @@ module Refile
         Refile.verify_uploadable(uploadable, @max_size)
 
         id = @hasher.hash(uploadable)
-        IO.copy_stream(uploadable, path(id))
-
+        begin
+          IO.copy_stream(uploadable, path(id))
+        rescue IOError
+          uploadable.open
+          retry
+        end
         Refile::File.new(self, id)
       end
 
--- a/refile.gemspec
+++ b/refile.gemspec
@@ -33,9 +33,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "aws-sdk"
   spec.add_development_dependency "rack-test", "~> 0.6.2"
   spec.add_development_dependency "rails", "~> 4.1.8"
-  spec.add_development_dependency "sqlite3"
+  spec.add_development_dependency RUBY_PLATFORM == "java" ? "activerecord-jdbcsqlite3-adapter" : "sqlite3"
   spec.add_development_dependency "selenium-webdriver"
   spec.add_development_dependency "yard"
   spec.add_development_dependency "rubocop"
-  spec.add_development_dependency "redcarpet"
 end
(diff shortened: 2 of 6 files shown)


lib/refile/backend/file_system.rb
```diff
@@ -34,8 +34,12 @@ def upload(uploadable)
         Refile.verify_uploadable(uploadable, @max_size)
 
         id = @hasher.hash(uploadable)
-        IO.copy_stream(uploadable, path(id))
-
+        begin
+          IO.copy_stream(uploadable, path(id))
+        rescue IOError
+          uploadable.open
+          retry
```
jnicklas: This seems super, super evil. Hello infinite IO loop. Why would this be necessary? Also, `uploadable.open` is not part of our API contract with regards to uploadables, no idea why it's even callable here.

refile.gemspec
```diff
@@ -34,9 +34,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "aws-sdk"
   spec.add_development_dependency "rack-test", "~> 0.6.2"
   spec.add_development_dependency "rails", "~> 4.1.8"
-  spec.add_development_dependency "sqlite3"
+  spec.add_development_dependency RUBY_PLATFORM == "java" ? "activerecord-jdbcsqlite3-adapter" : "sqlite3"
```
jnicklas: This does not work as you intend it to. This will work fine when developing, but when the gem is generated, the gemspec will depend on the platform it was *built on*, not that it is *run on*. This isn't a huge deal with dev-dependencies, but it's still weird.

Maybe we should just move all dev dependencies into the Gemfile.

spec/refile/spec_helper.rb
```diff
@@ -38,7 +38,8 @@ def presign
   File.open(tempfile.path, "r")
 end
 
-Refile.processor(:convert_case) do |file, format:|
+Refile.processor(:convert_case) do |file, options = {}|
+  format = options.fetch :format
```
jnicklas: Why does JRuby not support this syntax? Don't they use the same parser? Weird.

refile.gemspec
hashrocketeer: Good idea! That's typically how we handle dev dependencies. Bundler has more robust support for multiple platform dependencies.

spec/refile/spec_helper.rb
hashrocketeer: Pretty sure they have a different parser all together. For some reason the format was always coming through as nil. This is likely one of the things that will be ironed out before JRuby 9k drops, but doesn't work today.

lib/refile/backend/file_system.rb
hashrocketeer: Good point. We'll investigate further. When uploadable is a Tempfile, it always comes through as closed.

spec/refile/spec_helper.rb
jnicklas: This actually seems to be a bug in JRuby. Filed: https://github.com/jruby/jruby/issues/2518

janko-m:
Closing in favor of #187.

PR #94, opened 2015-01-07
Detailed README.md javascript library [ci skip]

asiniy:
There is step 3 :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,13 @@ Now mark the field for direct upload:
 <%= form.attachment_field :profile_image, direct: true %>
 ```
 
-There is no step 3 ;)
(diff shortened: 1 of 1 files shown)


asiniy:
Else one question.

I want to develop to the sources of refile. What can I do?
@jnicklas

jnicklas:
Permitting `cache_id` is no longer necessary on master, which is why it is no longer mentioned in the README. If you read the docs at [RubyDoc](http://www.rubydoc.info/gems/refile), you can get docs for the latest gem release. There are some issues with the current master, but hopefully we'll have it ironed out in the next couple of days.

As for helping out, just take a look at the open issues and see if there is anything you can fix :)

PR #97, opened 2015-01-08
bug on change file

lazaronixon:
when upload of file is completed refile [remove name attribute from metadataField](https://github.com/elabs/refile/blob/master/app/assets/javascripts/refile.js#L42) causing a bug when other file is selected.


jnicklas:
Huh. This is a bit dumb.

I don't really like this solution though. Ideally ids should be unique, so there really *shouldn't* be two fields with the same ID. Maybe we should just go back to `previousSibling` like we did before?

PR #99, opened 2015-01-09
Internazionalization

asiniy:
It's time to refile to be more sexy :)

I'm Russian native speaker and I know deutsch.

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -1,0 +1,8 @@
+de:
+  activerecord:
+    errors:
+      messages:
+        too_large: "Datei ist zu gross"
+        download_failed: "es gibt ein Fehler während Dateiladen"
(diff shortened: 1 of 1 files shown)


jnicklas:
@asiniy development can't be reliant on that. Imagine if we add a dozen languages. We now need to add a new translation. Should I keep a list of who the twelve people are who added the languages, email them and wait for each of their responses. What if one of them doesn't respond for whatever reason? I'm sure you can see how problematic that is, and what kind of burden it puts on me as maintainer.

Other gems don't do this either. Devise only has english translations, and has a list of translations to other languages in their wiki. This puts the burden on those who include the translation files, not on the maintainers of the gem.

asiniy:
@jnicklas you're right. Sorry for wasting your time

jnicklas:
@asiniy absolutely no worries :)

PR #101, opened 2015-01-09
Specify content type for S3#upload & App#stream_file.

LuckyThirteen:


jnicklas:
The idea is for the content type to be inferred from the extension. Stores should not be aware of file metadata.

PR #109, opened 2015-01-12
fix store metadata migration in readme

nhattan:


jnicklas:
Thank you for pointing out this problem, the PR you provided wasn't entirely correct either, but I went ahead and pushed a fix.

Issue #91, opened 2015-01-02
Setting custom headers when downloading files


hubertlepicki:
@jnicklas what's your view on how that should be implemented? I need to set content-disposition (have two buttons: "Open" and "Download" and they should do different thing for the same files.

I could set it up via a rack middleware independently from Refile, I guess, but maybe it's something the app should do?

jnicklas:
After a bit more reflection, i'm going to go with a no on this one.

PR #119, opened 2015-01-17
Add disposition option to attachment_url for forcing file downloads

### LINKED ISSUE #118: Creating a forced download link
This might be a super basic question but it's late and I'm stumped. I'm using Refile to serve up audio files uploaded by musicians:

``` ruby
class Song < ActiveRecord::Base
  attachment :audio_file
end
```

I'm also storing the _filename _size _content_type rows in my songs table

I want to create a link so that when someone clicks it, it automatically downloads. Currently, I can get it to open streaming via attachment_url, but not the equivalent of "save target as". I'm not sure how to serve this out from the controller in a send_file, either.

``` ruby
#View
  <%= link_to "download", attachment_url(song, :audio_file, filename: song.audio_file_filename)%>  #opens up an mp3 player
  <%= link_to "controller download ", user_album_song_download_path(current_user, @album, song) %> 

#controller - not sure what to do here, either
  def download
    send_file Refile.attachment_url(@song, :audio_file), type: @song.audio_file_content_type, disposition: :attachment
  end
```

Where am I going wrong, and is this in the documentation anywhere?

excid3:
Add option for requesting a file with the `disposition: "attachment"`.

This fixes #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,12 @@ no file has been uploaded:
 <%= attachment_image_tag(@user, :profile_image, :fill, 300, 300, fallback: "default.png") %>
 ```
 
+You can also force downloading of a file instead of rendering inline:
+
+``` erb
+<%= attachment_image_tag(@user, :cv, disposition: :attachment) %>
+```
+
 ## 5. JavaScript library
 
 Refile's JavaScript library is small but powerful.
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -248,7 +248,7 @@ module Refile
     # @param [String, nil] host            Override the host
     # @param [String, nil] prefix          Adds a prefix to the URL if the application is not mounted at root
     # @return [String, nil]                The generated URL
-    def attachment_url(object, name, *args, prefix: nil, filename: nil, format: nil, host: nil)
+    def attachment_url(object, name, *args, prefix: nil, filename: nil, format: nil, host: nil, disposition: nil)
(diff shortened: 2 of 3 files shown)


jnicklas:
Using query strings for this is a bit problematic, since we heavily encourage use of CDNs. CloudFront has query strings disabled by default, for performance reasons, some CDNs might not support query strings at all.

excid3:
Oh dang. Good point. What's a better way to solve this?

jnicklas:
See also #91, btw.

The solution is to put it into the URL. But it's *another* thing we need to put there. Since there are a number of people asking for this, we should fix this somehow. IMO it would make the most sense to put this at the beginning of the URL. As in `/inline/cache/12345/test.jpg` or something like that. We should let `Refile.attachment_url` fill this in as `inline` by default, so it doesn't have to be specified when we don't care about it.

I'm a bit torn on the name though. `Content-Disposition: inline` makes sense, but the word `inline` in of itself doesn't really communicate what it's trying to do.

excid3:
Well, I just found the `download` HTML5 attribute that makes links download the request instead of navigating to it. I think we can safely skip adding a disposition option into Refile and just recommend using that.

I can submit a PR for a Readme update if you want.

jnicklas:
Browser support is a bit shaky, it seems.

jnicklas:
Closing this PR for now. Let's move discussion back to #118.

Issue #106, opened 2015-01-11
Extract MiniMagick processing into a separate gem.


jnicklas:
People have to add a dependency on MiniMagick in their Gemfile anyway.

Putting this out there as a point of discussion. Should we do this? Why, why not?

kwando:
I think it makes sense to have it in separate gem.

balexand:
I would like to see it broken out into a separate gem. That way, another processing service could be substituted. For example, a service like http://www.imgix.com/ could be used. Or a microservice written in another language could be used.

jnicklas:
Done.

PR #125, opened 2015-01-22
Require mini_magick >= 4.0.0

kirs:
Otherwise using refile with mini_magick 3.8.1 causes `format` errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Features:
 Add the gem:
 
 ``` ruby
-gem "mini_magick"
+gem "mini_magick", "~> 4.0.0"
 gem "refile", require: ["refile/rails", "refile/image_processing"]
 ```
 
--- a/lib/refile/image_processing.rb
+++ b/lib/refile/image_processing.rb
@@ -1,5 +1,6 @@
 require "refile"
 require "mini_magick"
+require "mini_magick/version"
 
 module Refile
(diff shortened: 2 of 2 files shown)


janko-m:
As a maintainer of MiniMagick, I definitely vote for this. MiniMagick 4 has *lot* of improvements in robustness, speed and memory usage. The only thing was that before, MiniMagick 4 was not supported by JRuby, but that has changed with the last JRuby version.

The check that you made won't work for MiniMagick <= 3.8.0, because then `MiniMagick.version` didn't exist, there was just `MiniMagick::VERSION`, and it was a `String`, not a `Gem::Version` ([I didn't want that change](https://github.com/minimagick/minimagick/pull/233), but the maintainer just pulled it in anyways). I suggest the following (someone correct me if I'm wrong):

```rb
require "refile"
begin
  gem "mini_magick", ">= 4.0.0"
  require "mini_magick"
rescue Gem::LoadError => error
  raise error, "Refile requires mini_magick >= 4.0.0"
end
```

jnicklas:
I want to do #106, which would make this unnecessary. If we have a separate gem, it could  enforce a proper gem dependency.

jnicklas:
Since #106 is complete and the new refile-mini_magick gem will require mini_magick ~> 4.0 this is complete.

PR #126, opened 2015-01-23
Extract backend methods

rossta:
Consider this use case for Refile:

App using Refile also supports UGC markdown with images. Markdown preprocessor in app rewrites image srcs to Refile attachment URLs to a "remote url" to leverage processing, CDN and prevent mixed content warning for non-ssl images. 

Image srcs are `Base64` encoded in attachment urls to resolve to a proper route in `Refile::App`; they must be decoded by the backend for remote retrieval.

Example:
```markdown
![lolcat](http://example.com/lolcat.gif)

![lolcat](//refile-cdn.com/remote/aHR0cDovL2V4YW1wbGUuY29tL2xvbGNhdC5naWY=/processed.gif)
```

To support this, I've extracted a few methods that allow backends to customize the default behavior of the `verify_id` macro. These changes would allow our example remote url backend to decode the id and supply its own matcher to verify the id.

### DIFF
--- a/lib/refile/backend_macros.rb
+++ b/lib/refile/backend_macros.rb
@@ -7,7 +7,7 @@ module Refile
       mod = Module.new do
         define_method(method) do |id|
           id = id.to_s
-          if id =~ /\A[a-z0-9]+\z/i
+          if self.class.valid_id?(id)
             super(id)
           else
             raise Refile::InvalidID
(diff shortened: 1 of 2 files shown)


jnicklas:
I think more pressing than this is letting backends change what constitutes an invalid ID in the test suite. But this is still a pretty nice change, so I'm :+1: on merging this.

Please be careful in that Base64 encoding can contain the `/` character which could lead to path traversal vulnerabilities, or simply buggy behaviour if you're not careful. You'll need to URI escape the Base64 data before passing it through the URL like this.

rossta:
Cool, thanks for the feedback.

PR #130, opened 2015-01-26
Don't care if no record. Possible when we have hm/ho :picture association

route:


janko-m:
Sorry for not looking at this sooner. Could you provide an example where it happens that `nil` is passed in to the `#attachment_image_tag`?

route:
I think it's common to have dedicated Picture model like: Product has many Pictures or has one Picture which could be not uploaded yet. For such products we'd like to show default picture or fallback in terms of `attachment_image_tag`. So basically the code looks:
```
= attachment_image_tag(product.picture, :image, :fit, 110, 110, fallback: 'default.jpg')
```

janko-m:
Make sense! @jnicklas can we merge this?

jnicklas:
I agree with this PR in principle, but the implementation is a bit strange. The code just doesn't read very clearly. Especially since we currently don't have any unit tests for this method, that makes me a bit apprehensive.

janko-m:
I think if you just moved the `classes` line inside the `if file`, and then you can remove the conditional, it should really be fine without tests.

jnicklas:
But we use `classes` even if there is no file.

janko-m:
Lol, didn't see it properly. I think this is a really small change, so I think we could let it go without tests. @route, I just think in the second line you should write `if record` instead of `if file`, to keep the existing functionality (when record has no attachment, that it still gets the class by its name). Otherwise it reads clear to me.

route:
@janko-m Agree, rebased/updated

route:
Guys is it ok or should I refactor the whole method somehow?

janko-m:
I was waiting for @jnicklas to approve, but maybe he's busy, and this is a really small change, so I'm merging it. Thanks :)

PR #140, opened 2015-02-02
Fix README ruby-style

nyaa:
* Replaced double-quoted strings without interpolation with single ones

https://github.com/bbatsov/ruby-style-guide#strings

* Replaced `[]` arrays with space delimited array - `%w()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Add the gem:
 
 ``` ruby
 gem "mini_magick"
-gem "refile", require: ["refile/rails", "refile/image_processing"]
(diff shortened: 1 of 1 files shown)


rossta:
Based on the `.rubocop.yml` configuration, it appears that double quotes and the `%w[ ]` syntax are preferred in this project, though it looks like the `Gemfile` is not included in the rubocop spec.

jnicklas:
The authors of a project get to decide on its style. Deal with it. Please don't send this kind of PR, it's pretty obnoxious.

PR #152, opened 2015-02-13
Updated Readme for Direct Uploads

Austio:
This would have helped me when i first used the gem.  May be useful information for someone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,8 +266,16 @@ work.
 
 ### Direct Uploads
 
-By default you can directly upload to the refile app cache by sending a POST request to the uri
-'/attachments/cache' with a parameter {file: 'yourfile.jpg'}  
+You can directly upload to the refile app cache by sending a request to the uri
+```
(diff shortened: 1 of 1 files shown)


jnicklas:
I think the format of this is a bit of a mess. Content-Type is a header, not a param, and I have no idea what the `params` section is supposed to be. Maybe describing this in text would actually be more helpful, since the file can actually be sent encoded in a number of ways, just provided Sinatra can parse it.

Also, if we add a section like this, it should really talk about [`Refile.direct_upload`](http://www.rubydoc.info/gems/refile/Refile#direct_upload-class_method).

Austio:
Cool, will reformat and resubmit.  The content-type was part of the headers used in a successful send using a rest api client so i can see how that would be confusing to people just looking at it from a params standpoint.  

Is the location in the read me good or do you want a new section?

jnicklas:
I'll reopen once this receives further updates.

PR #162, opened 2015-02-24
Backend should know file content type

div:
@jnicklas Hey Jonas, thanks for a great gem!

The only thing I dislike right now, is the need to manually take care of Content-Type. What if backend could figure it out for us? We could store it in S3 or grab from the file system without the need to save it to the database or provide in the helper options. Here is the simplest code to make it work, just a proof of concept. What do you think?

### DIFF
--- a/lib/refile/app.rb
+++ b/lib/refile/app.rb
@@ -102,8 +102,7 @@ module Refile
       end
 
       filename = request.path.split("/").last
-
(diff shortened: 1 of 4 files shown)


jnicklas:
Sorry, I'm being harsh on your PRs and issues @div ;) I don't like this either :(

I'd rather keep backends as simple as possible. I think the current system works just fine. Storing it in the DB is just more flexible, and probably for many backends, actually faster.

PR #163, opened 2015-02-24
Reusing cache id on store!

div:
Can we use same ids when moving from cache to store?

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -70,6 +70,8 @@ module Refile
     def set(value)
       if value.is_a?(String)
         retrieve!(value)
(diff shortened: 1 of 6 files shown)


jnicklas:
I really dislike this. IMO the point of stores is that they are very opaque.

Could you please explain why you want this? Maybe understanding the use-case would help find a solution.

jnicklas:
I'm gonna say no to this one. @janko-m what do you think?

janko-m:
Well, I'm not really against this, but I don't understand why this would be useful. It imposes the rule that every backend can upload an image into a certain ID, which I think current backends all *can* do, but I still wouldn't like to impose it without a reason.

PR #177, opened 2015-03-17
Add ActiveRecord test for replacing files

janko-m:
Assignment of the attachment attribute caches the attachment, but doesn't change any actual columns on the model. Sequel, for example, won't save the record if no columns were changed, thus when replacing an attachment on an existing record it won't trigger the `before_save` callback. So, I thought it's good to have this check for ActiveRecord, just in case.


janko-m:
I'm not sure if then the tests for `remove_attachment` and `attachment_remote_url` should also be added for consistency, but I think this should be good enough.

jnicklas:
Nice, thank you!

PR #184, opened 2015-03-23
Initialize values in method signature instead of using the ||= operator.

tomciopp:
Cleans up the attachment_url method a bit.

### DIFF
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -235,13 +235,11 @@ module Refile
     # @param [String, nil] host            Override the host
     # @param [String, nil] prefix          Adds a prefix to the URL if the application is not mounted at root
     # @return [String, nil]                The generated URL
-    def attachment_url(object, name, *args, prefix: nil, filename: nil, format: nil, host: nil)
+    def attachment_url(object, name, *args, prefix: Refile.mount_point, filename: nil, format: nil, host: Refile.host)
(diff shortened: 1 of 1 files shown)


jnicklas:
Note that this is subtly not the same thing, and I seem to recall that I did it this way on purpose, though I can't quite remember why. It's different when the argument is explictly given as `nil`. In the case of a defaulted formal parameter, the actual argument will be `nil`. In the case of the `||=` it will still be the default value, and will not be `nil`. IIRC I wanted that behaviour, but I can't recall why.

tomciopp:
The only difference I can see is if you wanted to explicitly call this method with a falsey value as one of the arguments, which would then set the default value in the body of the method. (In the change the value would remain falsey) I could be wrong, but if that is the case I would argue in favor of the change, as that feels like a code smell to me as it feels awkward to change the parameter of a function mid method.

janko-m:
True. I thought at first that this was made to try to keep the method signature line short (as it's already over 80 columns as it is). But if that doesn't matter, I like the default arguments in the method definition more.

jnicklas:
I just remember we do it this way: previously, the Rails helpers actually set a default value of their own. This way, the default values would kind of cascade. Since d842f0f83e4b63910c5bbf19a885a2a326dfb8eb we don't do that anymore, so there's no point in differentiating the cases.

PR #193, opened 2015-04-01
Add max_size directly to attachment

lserman:
I know that `max_size` is already an option on the backend, but in my use-case I wanted two attachments to have different file size validations but be uploaded to the same backend. I think that it can be a pretty common use case (for me, I have one image and one video attachment on the same model).

Also, I thought it was a little weird that the `max_size` option is available only on the backend, while `content_type` and `extension` validations were available on the model.

After this commit I feel like there is a lot of validation code in the attacher, maybe some kind of `AttachmentValidator` class should exist (separate PR)?

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -3,12 +3,12 @@ require "open-uri"
 module Refile
   # @api private
   class Attacher
-    attr_reader :record, :name, :cache, :store, :options, :errors, :type, :valid_extensions, :valid_content_types
+    attr_reader :record, :name, :cache, :store, :options, :errors, :type, :valid_extensions, :valid_content_types, :max_size
     attr_accessor :remove
 
     Presence = ->(val) { val if val != "" }
 
-    def initialize(record, name, cache:, store:, raise_errors: true, type: nil, extension: nil, content_type: nil)
+    def initialize(record, name, cache:, store:, raise_errors: true, type: nil, extension: nil, content_type: nil, max_size: nil)
       @record = record
       @name = name
       @raise_errors = raise_errors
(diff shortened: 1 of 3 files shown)


jnicklas:
The reason that `max_size` is set on backends is because it is otherwise only an advisory validation. The validations for content type and filename are only there to provide nicer errors to users, they don't actually protect from malicious intent. I don't really see the point of doing the same thing with `max_size`, its only purpose is protecting the application from files which are too big to handle, if it's set in the model it cannot do that, so it becomes useless.

What you should do instead is create separate backends for the video and image files with different size restrictions. Otherwise someone could just circumvent it anyway.

lserman:
I think it is common to have a label on the form that says "max size 5mb" or whatever, and I just wanted a nice message to go along with that. I ended up just making a custom validation, but like I said I don't see this as uncommon and I will probably be making this same custom validation on every app that I make.

I think `max_size` protects the application from files which are too big just the same way `content_type` protects the application from files which are the wrong type. There is always a fallback, your application will have to enforce the real type and handle errors just as the backend will have to enforce the real size. Aren't both validations really only for user convenience? What's the difference?

jnicklas:
The difference is that it's very confusing to have two *different* ways of validating max_size, only one of which actually works the way one might expect. There should be one way of doing this, and that one way should be secure. Adding this to the model is simply superfluous.

> I think max_size protects the application from files which are too big just the same way content_type protects the application from files which are the wrong type.

That's the problem though, *neither* of these would actually protect from anything. In the case of content_type we are very clear on the fact that this is only for convenience, and that it doesn't actually protect anything.

PR #194, opened 2015-04-03
Option to only use cache

kobaltz:
Within my application, I've opted to exclude mounting the Sinatra application provided by Refile. However, this posed a problem when I tried to use the javascript callbacks when trying to access the cache to create a progress bar. I found that the implementation required to add this functionality, outside of including the Sinatra application, began creating obtrusive javascript. By adding an option to still include the Sinatra application, but optionally excluding the routes, I was able to accomplish this.

My goal was to protect the file uploads so that only authenticated users could access the files; authenticated logic handled separately. This pull request is fairly unobtrusive to the existing code base and adds what I consider to be important functionality.

### DIFF
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -74,6 +74,15 @@ module Refile
     # @return [Boolean]
     attr_accessor :automount
 
+    # Using your own custom routes, but still need access to the cache?
+    #
+    # If set to true then make sure automount is set to true as well. 
+    #
+    # The default is false.
+    #
+    # @return [Boolean]
+    attr_accessor :cacheonly
+
     # Value for generating signed attachment urls to protect from DoS
     #
     # @return [String]
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -81,7 +81,7 @@ module Refile
     # The default is false.
     #
     # @return [Boolean]
-    attr_accessor :cacheonly
+    attr_accessor :cache_only
 
     # Value for generating signed attachment urls to protect from DoS
     #
@@ -322,7 +322,7 @@ Refile.configure do |config|
   config.logger = Logger.new(STDOUT)
   config.mount_point = "attachments"
   config.automount = true
-  config.cacheonly = false
+  config.cache_only = false
   config.content_max_age = 60 * 60 * 24 * 365
   config.types[:image] = Refile::Type.new(:image, content_type: %w[image/jpeg image/gif image/png])
 end
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -313,6 +313,7 @@ Refile.configure do |config|
   config.logger = Logger.new(STDOUT)
   config.mount_point = "attachments"
   config.automount = true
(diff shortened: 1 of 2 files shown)


janko-m:
Sorry for the late response. If some URLs should be authenticated, couldn't you maybe generate the image URLs only to the authenticated user? Because the new `secret_key` already prevents unwanted people from accessing the images.

jnicklas:
I rather dislike how this is implemented. Setting this in a static context is bound to lead to problems. If we add this kind of option, I'd rather it work similarly to `Refile.direct_upload`.

kobaltz:
@janko-m This would work for certain applications. However, the scope of my project determines that authenticated and authorized users should be able to download the file. For example, if clients upload medical history forms, these forms should only be accessible to the appropriate people at a doctor's office. Not all people should be able to download the files. In a multi-tenant environment, other doctors offices should not be able to pull the medical records. Part of the HIPA Compliance would place restrictions where even a secret key wouldn't be enough to protect the user's file as it is still technically possible to download the information regardless of how improbable (thinking of recent breaches like the Anthem incident).

@jnicklas I'll look into how the `direct_upload` is working. I do not disagree that this wasn't a monkey patch, but it did patch an immediate issue which otherwise would disqualify Refile as an option.

janko-m:
@kobaltz How do you add those routes back, then? I think it's impossible to guess the token for the route, because each route has a different token generated (because it depends on the *path string* as well as the secret key). If people don't have permissions, and you don't generate the URL for them, I don't see how they can get to that.

Take S3 for example. A lot of people store confidential files on S3, and it's backed by the AWS key and secret. Theoretically someone could guess those keys as well. Isn't it the same?

kobaltz:
@janko-m I use a custom controller action which will for example

      def download
        @company = Company.friendly.find(params[:company_id])
        @user = @company.users.friendly.find(params[:user_id])
        authorize @user
        send_file @user.resume.download, disposition: :inline
      end

So when you call `@user.resume.download`, refile will generate a TempFile that you can then use `send_file` to download the file. Using this method, you still have access to your application's authorization logic. In this example, I'm using Pundit to authorize the `@user` record. If the current user is allowed to `download?` on the specified record, it proceeds with downloading the file.

This method is also not limited to simple file downloads. If your user had a `profile_image` attribute which stored the Refile information for the user's image, you could pass this through some post processing to resize the image. This does process the image on the fly, so you may want to have a CDN in place as suggested in the Refile documentation.

      def download
        @company = current_user.company
        @user = @company.users.friendly.find(params[:user_id])
        authorize @user
        processor = Refile.processor(:fill, Refile::ImageProcessor.new(:fill))
        temp_file = Tempfile.new('profile_image')
        temp_file.binmode
        temp_file.write @user.profile_image.read
        temp_file.rewind
        image_file = MiniMagick::Image.new(temp_file.path)
        file = processor.fill(image_file, 150, 150)
        temp_file.close
        send_file @user.profile_image.download, disposition: :inline
      end

Issue #142, opened 2015-02-05
Doesn't work on nested resources


alexandremcosta:
Does refile works on nested resources? I don't think so.
I'm able to update every single normal field, but not the refile images.

One detail is that if I debug, I can see that `@page.update_attributes(permitted_params)` does NOT update refile images, so `@page.studio.profile_image_id` is still nil after that. But `@page.studio.profile_image` actually returns a refile object during the debugging section. After the debugging, the object is not there anymore, neither profile_image_id is set.

And if I do something like `@page.studio.update_attributes(permitted_params[:studio_attributes])`, during the debugging section, it does persist the image and the profile_image_id!

Here is the code:

page.rb
```ruby
class Page < ActiveRecord::Base
  has_one :studio
  accepts_nested_attributes_for :studio
end
```

studio.rb
```ruby
class Studio < ActiveRecord::Base
  belongs_to :page

  attachment :profile_image
  attachment :background_image
end 
```

pages_controller.rb
```ruby
PagesController < ApplicationController
  prepend_before_filter :load_object, only: [:edit, :update]

  def edit
  end 

  def update
    if @page.update_attributes(permitted_params)
      redirect_to account_path
    else
      render 'edit'
    end 
  end 

  private
  def load_object
    @page = Page.find(params[:id])
    @page.build_studio if @page.studio.nil?
  end 
  def permitted_params
    params.require(:page).permit(
      :slug,
      studio_attributes: [:id, :name, :profile_image, :remove_profile_image, :background_image, :remove_background_image])   
  end 
end
```

views/pages/edit.html.erb
```erb
<%= form_for @page do |f| %>
  <%= f.fields_for :studio do |studio_form| %>
    <p class="form-group">
      <%= studio_form.label :name, t('studio.name') %>
      <%= studio_form.text_field :name, :class => 'form-control' %>
    </p>
    <div class="form-group">
      <%= studio_form.label :profile_image, t(:profile_image) %>
      <% if studio_form.object.profile_image_id? %>
        <div class="image">
          <%= attachment_image_tag(studio_form.object, :profile_image, :fill, 100, 100) %>
          <%= studio_form.check_box :remove_profile_image %>
          <%= studio_form.label :remove_profile_image %>
        </div>
      <% end %>
      <%= studio_form.attachment_field :profile_image, direct: true, presigned: true, :class => 'form-control' %>
    </div>
    <div class="form-group">
      <%= studio_form.label :background_image, t(:background_image) %>
[...]

PR #201, opened 2015-04-17
Nested attachments

aglushkov:
When updating attachments in nested models, it [fails](https://github.com/refile/refile/issues/142)
This is a fix

### DIFF
--- a/lib/refile/attachment/active_record.rb
+++ b/lib/refile/attachment/active_record.rb
@@ -21,6 +21,11 @@ module Refile
           end
         end
 
+        define_method "#{name}=" do |value|
+          send("#{name}_id_will_change!")
+          super(value)
+        end
+
         before_save do
           send(attacher).store!
         end
--- a/spec/refile/active_record_helper.rb
+++ b/spec/refile/active_record_helper.rb
@@ -11,6 +11,7 @@ ActiveRecord::Base.establish_connection(
 class TestMigration < ActiveRecord::Migration
   def self.up
     create_table :posts, force: true do |t|
+      t.integer :user_id
       t.column :title, :string
       t.column :image_id, :string
       t.column :document_id, :string
@@ -18,6 +19,10 @@ class TestMigration < ActiveRecord::Migration
       t.column :document_content_type, :string
       t.column :document_size, :integer
     end
+
+    create_table :users, force: true do |t|
+    end
+
   end
 end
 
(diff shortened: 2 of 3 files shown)


janko-m:
This looks good. I didn't know ActiveRecord acted in that way, I thought it's always going to save the records, regardless of whether they're dirty or not. But I guess it works differently for nested models.

@jnicklas, merge?

jnicklas:
I'm not a huge fan of how the tests are written. Defining constants in before blocks is very bad practice, imo, and bound to lead to tears later. I also really dislike inlining migrations like this. Migrations are slow, we should ideally not add and remove tables at runtime in the suite.

janko-m:
Yeah, originally I thought to just merge (as accepting the idea), and then correct the tests, but we might forget and it's better that tests are right from the beginning.

@aglushkov You could reuse the existing "posts" table (see active_record_helper.rb), and add another table there. Then you can define the new nested model with `Class.new` as was done in the before block on the top of the test file.

aglushkov:
I found that this solution only works for has_one association. I'll add more tests and rewrite this pull request

janko-m:
Perfect, good job!

Before I also wanted to add this "will_change" behaviour, because e.g. the Sequel ORM only saves records when something changed (or you tell Sequel that it's going to change), so I thought that ActiveRecord might too, but ActiveRecod always saves the record. Except with has_one nested attributes.

jnicklas:
Nice work @aglushkov. Tests look much better now :)

Weird question though: This assumes that `#{name}_id_will_change!` always is defined, can we make that assumption? It is currently possible, though as @janko-m noted: really weird, to define an attachment on a model without an ID column. Should we always assume there's an ID column. Should we always assume this column is a normal AR column?

janko-m:
Well, if the ID column isn't a column (e.g. it's an attr_accessor), I don't think that use case would have made sense. Because users should always want to persist the image (especially on an AR model), otherwise it gets lost. Also I think in that case they were much better off using a PORO instead of an AR model.

One small downside of this functionality is that, if users forgot the create an ID column, they will get an `undefined method #{name}_id_will_change!` error instead of `undefined method #{name}`. But I think that users will rarely forget that, and they can still figure out from the first message what's going on.

jnicklas:
Yeah, I think as you suggested somewhere else, we should just make it a hard error if there's no `id` column. Like it used to be.

Issue #135, opened 2015-01-28
Simplify interface of attachers towards helpers


jnicklas:
Currently the helpers have to dig into attachers in order to make all of their magic work. Is there another way? Can we make this interface lighter and easier? This comes up for example when trying to implement #6. In order for something that is not a true attachment to work with the helpers, we have to implement a lot of pretty weird, non-obvious stuff. Ideally this interface should be clean and simple. But how?

PR #202, opened 2015-04-21
Split part of Attacher off into AttachmentDefinition

jnicklas:
This splits Attacher into two classes, Attacher and AttachmentDefinition. AttachmentDefinition contains all the stuff from attacher that we can know in a static context (that is without having an instance of a record with an attachment). This might seem pretty pointless, but it's a first step towards fixing #6, everyone's favourite issue. (See also #135).

The problem we need to solve to implement for #6 is that we currently use Attacher from the `attachment_field` helper, but this won't work in the future, because we won't have an instance of the record we're trying to build. With this PR we only use AttachmentDefinition for which we only need the class, and not an instance of it.

The disadvantage of this PR is that the naming is all screwed up. AttachmentDefinition is a pretty terrible name, IMO. As is Attacher really. Can we come up with better names for this stuff?

### DIFF
--- a/lib/refile/attachment_definition.rb
+++ b/lib/refile/attachment_definition.rb
@@ -1,0 +1,46 @@
+module Refile
+  # @api private
+  class AttachmentDefinition
+    attr_reader :record, :name, :cache, :store, :options, :type, :valid_extensions, :valid_content_types
+    attr_accessor :remove
+
+    def initialize(name, cache:, store:, raise_errors: true, type: nil, extension: nil, content_type: nil)
+      @name = name
+      @raise_errors = raise_errors
+      @cache_name = cache
+      @store_name = store
+      @type = type
+      @valid_extensions = [extension].flatten if extension
+      @valid_content_types = [content_type].flatten if content_type
+      @valid_content_types ||= Refile.types.fetch(type).content_type if type
+    end
+
+    def cache
+      Refile.backends.fetch(@cache_name.to_s)
+    end
+
+    def store
+      Refile.backends.fetch(@store_name.to_s)
+    end
+
+    def accept
+      if valid_content_types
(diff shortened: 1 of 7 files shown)


janko-m:
Definitely like any attachment refactoring. I will try to review it by the end of this week to get a better understanding of the idea (and see about the naming).

lib/refile/attachment_definition.rb
```diff
@@ -0,0 +1,46 @@
...
+    def raise_errors?
+      @raise_errors
+    end
+
+    def validate(attacher)
+      errors = []
+      errors << :invalid_extension if valid_extensions and not valid_extensions.include?(attacher.extension)
+      errors << :invalid_content_type if valid_content_types and not valid_content_types.include?(attacher.content_type)
+      errors << :too_large if cache.max_size and attacher.size and attacher.size >= cache.max_size
+      errors
```
jnicklas: I'm not sure whether this belongs here or in `Attacher`.

janko-m: I think this is good, I imagine that Attacher is interfacing with database columns, and AttachmentDefinition doesn't know about it. Validations themselves don't use the model to build errors, so it's good that it's in AttachmentDefinition.

janko-m:
I think the naming is really fine. I don't know about you, but it's still hard for me to keep all parts of Refile in my head; what should go where, what is responsible for what etc. No matter how you put it, file upload is complicated :P (Refile is just the simplest implementation of it). I think that when we start fixing #6, it's going to be clearer what we want, and we can just change the naming (`AttachmentDefinition` shouldn't be used by the user directly anyway). So, I'm :+1: for merging.

jnicklas:
Done!

PR #210, opened 2015-04-28
Fix uploading large files to S3 via remote_url method

aglushkov:
When I was trying to upload large image like [this one](https://markrahrens.files.wordpress.com/2014/03/dsc_1295.jpg).
I see this error IOError: closed stream

screenshot from 2015-04-28 16 18 04

Somehow RestClient return closed file when uploaded files are large ( ~ from 15 MB)
And S3 wants this file to be opened

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -102,6 +102,7 @@ module Refile
           content_type: response.headers[:content_type]
         }
         if valid?
+          response.file.open if response.file.closed? # https://github.com/refile/refile/pull/210
           @metadata[:id] = cache.upload(response.file).id
           write_metadata
         elsif @raise_errors


janko-m:
Looks great! I guess it would be difficult to write a test for this. Could you add a comment at the end of the line with a short explanation? Just that we don't accidentally break it while refactoring.

aglushkov:
Now I think, maybe better to move this to [Refile::Backend::S3#upload](https://github.com/refile/refile/blob/master/lib/refile/backend/s3.rb#L55) ?

janko-m:
Yeah, I agree, this is a better place. @jnicklas, ok for merging?

jnicklas:
I don't think we can assume that the uploadable responds to `open` and `closed?`, can we? Did you actually run the S3 tests? I'm assuming that they fail given this change.

aglushkov:
You are right, I didn't think about it. And I didn't run S3 tests. 

So what to do? 
1. Place this in Attacher, where we know for sure  that Tempfile responds to `open` and `closed?`.
2. Check here in S3 class that uploadable responds to  `open` and `closed?`

jnicklas:
@aglushkov personally I find option (1) better.

aglushkov:
Done

PR #212, opened 2015-05-01
[WIP] Add HEAD support and respond with ETag and Last-Modified headers

### LINKED ISSUE #67: Make HEAD requests efficient
It should be possible to perform a HEAD request to check if a file exists, without paying for download, processing, etc...

cristianbica:
Fixes #67
This still needs some touching (tests for the headers) but I would like some feedback. Thanks

Changes:
- [rails] mounting routes without :via it seems it doesn't catched HEAD requests; can't find the exact cause but probably something related to the rails router and `Rack::Head`
- added ETag and Last-Modified headers; other storage adapters will have some issues with this; probably if the adapter doesn't respond to the method we should read the file and calculate the ETag
- if it's a HEAD request halt as HEAD doesn't expects a reponse body

TODO:

### DIFF
--- a/lib/refile/backend/file_system.rb
+++ b/lib/refile/backend/file_system.rb
@@ -92,6 +92,22 @@ module Refile
         ::File.exist?(path(id))
       end
 
+      # Return eTag of the file
+      #
+      # @param [Sring] id           The id of the file
+      # @return [String]
+      verify_id def etag(id)
+        Digest::MD5.file(path(id)).hexdigest
+      end
+
+      # Return last modified time of the file
+      #
+      # @param [Sring] id           The id of the file
+      # @return [Time]
+      verify_id def last_modified(id)
+        open(id).mtime
+      end
+
       # Remove all files in this backend. You must confirm the deletion by
       # passing the symbol `:confirm` as an argument to this method.
       #
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,5 @@
 if Refile.automount
   Rails.application.routes.draw do
-    mount Refile.app, at: Refile.mount_point, as: :refile_app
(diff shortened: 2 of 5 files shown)


janko-m:
I don't know how much this really improves performance. For the file system backend it's ok, but for any other backend which are over HTTP (which is most of them) you have to make one HTTP request for both E-Tag and Last-Modified, just to check if they are modified. I thought that files can't get modified in Refile, that they can only be reuploaded to a different place (with different backend ID)?

lib/refile/backend/file_system.rb
```diff
@@ -92,6 +92,22 @@ def initialize(directory, max_size: nil, hasher: Refile::RandomHasher.new)
         ::File.exist?(path(id))
       end
 
+      # Return eTag of the file
+      #
+      # @param [Sring] id           The id of the file
+      # @return [String]
+      verify_id def etag(id)
+        Digest::MD5.file(path(id)).hexdigest
```
jnicklas: I think we should at least use SHA256. There is probably no way to exploit this, but just to be sure. MD5 is known to be broken. I can't imagine that the difference in speed is really going to make a difference.

jnicklas:
Agree with @janko-m's concerns here. Adding the etag and last_modified headers is not worth it, since the point is that files are immutable, they shouldn't ever be modified. Also there are no tests for any of this behaviour, which is pretty much a necessity, imo.

cristianbica:
So the reasoning for this was I had a software accessing some files attached with refile (without a http cache in front). The software would make a HEAD query to check if the file changed. My issue was fixed by adding a http cache in front of the refile app (the recommended way :) ). 
I now agree with you guys that this doesn't adds much value to refile so I'm going to close this PR.
Thanks for the great work on refile :+1:

PR #218, opened 2015-05-09
Upgrade to AWS SDK for Ruby - Version 2

lazaronixon:


spec/refile/spec_helper.rb
```diff
@@ -61,6 +61,10 @@ def read(*args)
       @io.read(*args)
     end
 
+    def rewind
+      @io.rewind
+    end
```
jnicklas: Why is this addition necessary?

lazaronixon: https://github.com/aws/aws-sdk-ruby/blob/0d054b6eaaf6653210562e242a8586226565cbf4/aws-sdk-core/lib/seahorse/client/param_validator.rb#L106

PR #220, opened 2015-05-10
fix attachment_tag no longer having cdn host

grosser:
https://github.com/refile/refile/commit/acb99c2f42efa596b9f2d9a3a5f124b5c2df049a

introduced a bug where attachment_tag would previously pass down nil to
attachment_url, but now this is no longer overwritten as Refile.host

also included a failing test :)

@jnicklas


jnicklas:
I think the test is a strange mix of integrated (using a real ActiveRecord object) and mocked (mocking the Rails parts). Personally I'd prefer a more integrated tests. We had some mock-heavy unit tests for the helper before and ended up removing them because they weren't helpful.

grosser:
all integration now, better ?

jnicklas:
Yes, nice, thanks!

PR #228, opened 2015-05-15
fix #227 uploads are not appended when sending in cached files

### LINKED ISSUE #227: using multi upload should not overwrite existing uploads
when a user already has a bunch of files and selects more the previous ones get removed ... which might make sense in some usecase, but I'd guess most of the time users add more files and don't overwrite all their uploads.

@jnicklas

grosser:
@jnicklas


jnicklas:
Nice catch, and nice PR! Thank you!

PR #247, opened 2015-05-27
Add instructions to README for using Refile in Rails Engines

cynektix:
I wanted to use Refile in a Rails Engine and ran into the problem, that I wanted to require `refile-mini_magick`. This lead to the following error which confused me a lot:

```
bin/rails:6: warning: already initialized constant APP_PATH
```

The only error was the wrong required file, so maybe this might help someone in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -714,6 +714,24 @@ end
 Refile will now fetch the file from the given URL, following redirects if
 needed.
 
+## Using Refile inside of Rails Engines
+
+If you want to use Refile inside of a Rails Engine you need to load it the folloing way:
+
+``` ruby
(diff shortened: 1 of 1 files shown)


jnicklas:
I'm not really sure how *else* you would require refile in a Rails engine (aside from the fact that you don't need to require mini_magick). This is also pretty much the same thing you'd do if you're not in an engine.

cynektix:
Ok well the only thing that I did wrong was that I used the wrong path for the refile mini magick gem. So maybe it's not that big deal. Thanks for the hint with mini_magick.

Issue #6, opened 2014-11-26
Support multiple file uploads


jnicklas:
Both for direct and for regular uploads.

cpowell:
I can confirm that the technique described in the following blog post about CarrierWave is adaptable to the Refile gem:

http://u.osu.edu/hasnan.1/2014/03/30/rails-4-multiple-file-upload-with-carrierwave-nested-form-and-jquery-file-upload/

(I speculate that this is what Piotr Usewicz was referring to.)

nhattan:
I'd tried the solution that @cpowell cited above before. It's good but the I've got a problem with presigning. My context is:
```ruby
version.rb
class Version < ActiveRecord::Base
  has_many :attach_files
end

attach_file.rb
class Attach  < ActiveRecord::Base
  attachment :file
  belongs_to :version
end
```
I'm trying to upload multiple files for a version in one form. It'd be awesome if each file is uploaded to fog (such as S3) without hitting app.

jnicklas:
I've had some time to think about this, and I think I have a pretty decent way forward for this. Basically along @hubertlepicki's original suggestion. Posting this here for critique.

We add a macro to ActiveRecord called `accepts_attachments_for`, to be used like this:

``` ruby
class Post < ActiveRecord::Base
  has_many :images

  accepts_attachments_for :images, attachment: :file
end
```

Where `attachment: :file` names the column in the `Image` model that has the attachment, the default being `:file`, so it could have been omitted in this case.

This would generate two methods on `Post`, called `images_files` and `images_files=`. These can then be used in a form like this:

``` ruby
form_for @post do |form|
  form.attachment_field :images_files, multiple: true, direct: true
end
```

If we want to get really fancy we could allow:

``` ruby
form_for @post do |form|
  form.attachment_field :images, multiple: :file, direct: true
end
```

But this might just be more confusing.

We build in support into the JS to pick up on the `multiple` attribute, and do the right thing. With the recent changes in how the caching works, it should be relatively simple to preserve form redisplay behaviour.

Obviously this assumes that `Image` has no required fields aside from `file`, but this is reasonable, IMO.

IMO this macro would be small and simple enough to include it in Refile itself, without putting it in another gem.

Thoughts?

pusewicz:
Sounds really good to me!

geetfun:
+1 for @jnicklas's proposal.

Any other attributes that may need to be edited on :images records could be handled by accepts_nested_attributes_for *after* the initial upload as well.

[...]

PR #249, opened 2015-05-27
Fix for form upload update eating attachments

UnderpantsGnome:
@jnicklas This is a possible fix for the issue I mentioned [here](https://github.com/refile/refile/issues/6#issuecomment-101465403), the tests are passing.

### DIFF
--- a/lib/refile/rails/attachment_helper.rb
+++ b/lib/refile/rails/attachment_helper.rb
@@ -83,11 +83,19 @@ module Refile
 
       hidden_field(object_name, method,
         multiple: options[:multiple],
-        value: object.send("#{method}_data").to_json,
+        value: hidden_field_value(object, method),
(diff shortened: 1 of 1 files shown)


jnicklas:
This is not the right place to fix this problem. This should really be fixed inside `data`, possibly with a `select(&:present?)` or something. Also a failing test for this would be really nice.

UnderpantsGnome:
@jnicklas I'll see if I can make some time to provide a better solution including test in the near future.

jnicklas:
@UnderpantsGnome had some time to work on a fix for this. Can you give it a go and see if it solves your problem?

UnderpantsGnome:
@jnicklas That change you made seems to have resolved the issue. Sorry about the delay, things have been a little busy.

PR #251, opened 2015-05-28
when available, use mime/types/columnar which is more memory efficient

midu:
I stumbled upon [this PR](https://github.com/thoughtbot/paperclip/pull/1873) on paperclip, by @schneems, and I thought that refile could use the change as well.

Looks like a decent amount of memory could be saved by using the Columnar version of mime-types when available, with little to no changes.

The changes on this PR were shamelessly taken from similar PRs by @jeremyevans (mikel/mail#880) & @schneems (thoughtbot/paperclip#1873)

- see: https://github.com/mime-types/ruby-mime-types/pull/96

### DIFF
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -3,7 +3,14 @@ require "fileutils"
 require "tempfile"
 require "rest_client"
 require "logger"
-require "mime/types"
+
+begin
+
+  # Use mime/types/columnar if available, for reduced memory usage
+  require "mime/types/columnar"
+rescue LoadError
+  require "mime/types"
+end
 
 module Refile
   # @api private


schneems:
:+1: thanks for the mention, @jeremeyevans was the original author of that PR to mail https://github.com/mikel/mail/pull/880

midu:
oops, sorry @jeremyevans!

jnicklas:
While I appreciate that this is a great memory saving step, shouldn't this really be fixed upstream? We shouldn't just be littering all gems with conditional requires. Is there some downside to the columnar mode? Why didn't they just make it the default?

jeremyevans:
@jnicklas Columnar mode will become the default starting in mime-types 3. See https://github.com/mime-types/ruby-mime-types/pull/96 for details.

midu:
My understanding is that in version 2, making the columnar (plain text) files the default, would conflict with overriding the file path to the mime types using an ENV variables (json file). So this is an opt-in feature until version 3.

That being said, that sounds likes a choice a user should be making, not a gem like refile. In fact this could even break a project that is relying on the ability to override the load path.

I **think** that as a user, doing something like this in their Gemfile would get the benefits of `columnar` across all my gems, and allows **me** to decide whether or not I want that feature.

```rb
gem "mime-types", "~> 2.6.1", require "mime/types/columnar"
```

I am not convinced this PR makes sense anymore.

jnicklas:
Yeah, I agree. This doesn't belong in a gem. I think the mime-types people should provide an interface to configure which mode to use rather than relying on require order (maybe they already do?).

schneems:
Most people aren't aware that they're using mime-types, they're only aware that they added "refile" to their gemfile, and don't know it depends on mime-types. Most people won't make that change to take advantage of the memory optimizations. On the flip side they will `bundle install` on a new system which would give them the most recent mime-types (if not already installed). We're trying to get popular gems to support it now so we're better prepared and better understand the larger issues that come up when we transition to making it the default. This change has already been merged into mail/fog/carrierwave. Is there a possibility it will cause problems? Yes, versioning is hard and software are hard. If people run into edge cases.

The flip to what you said is true as well. If someone knows that they're using a gem that uses the ENV var, they can manually force mime-types to a lower value.

```
gem "mime-types", "< 2.6"
```

Either approach requires knowledge of the use of `mime-types` which 99% of users don't have. In one case they would get an error message and bugs that would hopefully be easily looked up, in the other case they're silently using 20% more memory in their starting rails app that they need to be because they added this gem.

PR #252, opened 2015-05-29
add gorails video to readme documentation

sublimecoder:
Add GoRails, powered by One month, video to readme documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,10 @@ And start uploading! Finally show the file in your view:
 <%= image_tag attachment_url(@user, :profile_image, :fill, 300, 300, format: "jpg") %>
 ```
 
+### Video Tutorial
+Get started with an example usage of the refile gem.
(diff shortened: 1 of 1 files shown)


jnicklas:
That tutorial is very nice, but it's also a little out-of-date. Which is not surprising given that Refile is such a young gem and still changing very fast. I'm not sure we should be pushing for it that hard by putting it that high up in the README. Maybe I'm just paranoid, all the basics are still the same, but I'm not sure.

sublimecoder:
No worries, I can understand your caution. Such is the nature of anything public that people depend on. Maybe drop it towards the bottom of the readme with a disclaimer about how quickly refile's API is changing so the video may not be 100% accurate?

I might also be able to get Chris to do an updated video.

jnicklas:
Closing this, the video has not kept up-to-date and I don't feel comfortable linking to it.

PR #254, opened 2015-05-30
Fix #253 Temporary files must be flushed to disk before the webserver read them

### LINKED ISSUE #253: Webserver is sending half written tempfile
In certain conditions web server is sending tempfile content before it's fully flushed to disk. It results in broken images at bottom or incomplete files (broken pdf, ...)

I've made a patch to force filesystem sync.

Titinux:


jnicklas:
Lots of test failures. Can you take a look and see what's up with that?

jnicklas:
Also what OS/Ruby version are you experiencing this with?

Titinux:
My bad. I shouldn't make code at 3 am :sweat: 

I'm on gentoo / linux 4.0 with the 2.3.0-dev version of ruby. My /tmp is tmpfs

jnicklas:
Awesome, thanks!

PR #259, opened 2015-06-01
Remove mini_magick Gemfile line in quick start

habibalamin:
`refile-mini_magick` already has `mini_magick` as a dependency.


jnicklas:
Thanks!

PR #266, opened 2015-06-03
ship less files and simplify gemspec

grosser:
@jnicklas 

shipping a bunch of test / misc files means download takes longer and vendor/cache gets bigger

adding spec to require-path means the spec files are actually in the gems load path ... but they are not supposed to be required ... right ?

also removing LOAD_PATH modifications

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -78,3 +78,6 @@ Lint/DefEndAlignment:
 
 Lint/HandleExceptions:
   Enabled: false
+
+Style/SpecialGlobalVars:
+  Enabled: false
--- a/refile.gemspec
+++ b/refile.gemspec
@@ -1,7 +1,4 @@
-# coding: utf-8
-lib = File.expand_path("../lib", __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require "refile/version"
+require "./lib/refile/version"
 
 Gem::Specification.new do |spec|
   spec.name          = "refile"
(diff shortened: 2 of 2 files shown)


jnicklas:
The test files are actually included in the gem very deliberately. It's so that backend implementations can require the spec files and run Refile's test suite against their own implementation. See e.g. https://github.com/refile/refile-s3/blob/master/spec/refile/s3_spec.rb#L1

The rest of the Gemfile is just want bundler autogenerates.

I think it's certainly possible to slim down the gem a little bit (we don't need the bundled Rails app to run feature tests for example), but is this really that much of an issue?

grosser:
I like my gems as slim as possible ... I'll take another look ...

grosser:
updated, looks good ?

a good chunk of saving when just excluding test_app
`git ls-files lib spec Readme.md`.split($/).reject { |f|
f.include?("test_app") }.size
=> 38
>> `git ls-files lib spec Readme.md`.split($/).size
=> 62

and half of master:
git ls-files | wc -l
      79

jnicklas:
Currently fails due to annoying rubucop errors. Sorry.

grosser:
Is ignoring english variables ok ? :)

grosser:
@jnicklas good to go ?

grosser:
What do you think about moving the reused spec files into lib instead ?

jnicklas:
@grosser I chose not to do that, because of my bad experience with Capybara in that regard. Can't count  the number of times people didn't find the spec files because they were in `lib`. :(

PR #267, opened 2015-06-03
Fix #244: Response codes for exceptions

### LINKED ISSUE #244: Graceful uploading error handling
For now application [doesn't](https://github.com/refile/refile/blob/master/lib/refile/app.rb#L64) handle backend fail cases and just responds with 500 error code, which is not convenient. So maybe it would be nice to handle those exceptions in app and return 413 in case of size validation, or 400 when ArgumentError is thrown. It'd improve semantics and probably help handling errors on frontend.

toothfairy:
Handles uploading exceptions and responds with appropriate status

### DIFF
--- a/lib/refile/app.rb
+++ b/lib/refile/app.rb
@@ -85,7 +85,7 @@ module Refile
       "forbidden"
     end
 
-    error ArgumentError do
+    error Refile::InvalidFile do
       status 400
       "Upload failure error"
     end
--- a/lib/refile/app.rb
+++ b/lib/refile/app.rb
@@ -85,6 +85,11 @@ module Refile
       "forbidden"
     end
 
+    error ArgumentError do
+      status 400
+      "Upload failure error"
+    end
+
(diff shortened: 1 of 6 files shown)


lib/refile/app.rb
```diff
@@ -85,6 +85,16 @@ class App < Sinatra::Base
       "forbidden"
     end
 
+    error ArgumentError do
```
jnicklas: ArgumentError is a very general error to rescue. I think it's not ideal to assume that any particular thing happened given an ArgumentError.

toothfairy: @jnicklas should it raise different kind of exception (ie Refile::InvalidFile or UnreadableFile) in backend macros?

jnicklas: That's a good idea probably. Yes.

jnicklas:
Nice :) Thanks!

toothfairy:
:dancer:

PR #271, opened 2015-06-09
List refile-mongoid as an ORM gem

krettan:
List the [refile-mongoid](https://github.com/krettan/refile-mongoid) extension which implements the `Refile::Mongoid::Attachment` under "Other ORMs".

_Background:_ To use refile in a rails application with mongoid and no ActiveRecord, a gem like [refile-sequel](https://github.com/refile/refile-sequel) is needed. Looked like there was none, so I took the liberty of writing one. It's basically a clone of [refile-sequel](https://github.com/refile/refile-sequel) but using mongo instead.


jnicklas:
Nice :) Awesome!

PR #276, opened 2015-06-16
Fix attachment extension validation when extension is in randomcase

caulfield:
I found a bug when uploaded file extension name is randomcased. (Iphone creates movies with MOV format for example). And app hasn't content type validation, only extensions list.
I wrote some tests and fixed invalid behaviour.

### DIFF
--- a/lib/refile/attachment_definition.rb
+++ b/lib/refile/attachment_definition.rb
@@ -37,7 +37,8 @@ module Refile
 
     def validate(attacher)
       errors = []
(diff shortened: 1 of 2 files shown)


caulfield:
Oh. So fast. I just pushed tests and rubocop fixes..

jnicklas:
Ahh, awesome. I was a bit trigger happy and had resigned to fixing those myself :P

jnicklas:
@caulfield that would be great, thanks!

jnicklas:
Had a fix almost finished, so I just pushed it.

caulfield:
:+1:

PR #279, opened 2015-06-17
make the default instructions the same region as the previous default…

grosser:
… so images stay visible

previously the default region was used us-west-2
when upgrading all my images were gone because I did not know what region my bucket was in
I was able to find it via `Refile.store.instance_variable_get(:@bucket).location_constraint` but that took me some digging ...

@jnicklas

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ require "refile/s3"
 aws = {
   access_key_id: "xyz",
   secret_access_key: "abc",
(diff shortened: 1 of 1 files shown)


jnicklas:
Huh. Isn't the default us-east?

grosser:
yeah I think you are right ... our bucket was created via the aws console ... not sure how it worked without specifying the region ...

PR #281, opened 2015-06-19
Fix Encoding::UndefinedConversionError during upload (Related to #187)

scarfacedeb:
When you try to upload a binary file using jruby, it raises Encoding::UndefinedConversionError.

In related #187 issue @janko-m determined that it only happens when you use a wrapper around an IO (e.g. Refile::File).

I've found out that it happens due to `Encoding.default_internal = Encoding::UTF_8` line in rails internals.
see: https://github.com/jruby/jruby/issues/2779#issuecomment-113473546 for details

In this PR, I tried to workaround the issue by directly opening destination file with binary mode during upload.
It should make it clear for `IO.copy_stream` that dest file is in binary, ascii-8bit encoding and it doesn't need to convert the content of uploadable file to default internal encoding (which is UTF_8) prior to writing it to the destination.

I tested it with jruby-9.0.0.0.rc1.

### DIFF
--- a/lib/refile/backend/file_system.rb
+++ b/lib/refile/backend/file_system.rb
@@ -34,7 +34,7 @@ module Refile
       # @return [Refile::File]      The uploaded file
       verify_uploadable def upload(uploadable)
         id = @hasher.hash(uploadable)
-        IO.copy_stream(uploadable, path(id))
+        ::File.open(path(id), "wb") { |dest| IO.copy_stream(uploadable, dest) }
 
         Refile::File.new(self, id)
       end
--- a/spec/refile/backend_examples.rb
+++ b/spec/refile/backend_examples.rb
@@ -33,7 +33,7 @@ RSpec.shared_examples_for :backend do
       expect(retrieved.exists?).to be_truthy
     end
 
-    context 'when Encoding.default_internal is set to UTF-8' do
+    context "when Encoding.default_internal is set to UTF-8" do
       let(:binary) { "\x89PNG\r\n\x1A\n\x00\x00".force_encoding("ASCII-8BIT") }
       let(:binary_file) { uploadable(binary) }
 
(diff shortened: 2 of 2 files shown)


janko-m:
Thank you, @scarfacedeb!

lib/refile/backend/file_system.rb
```diff
@@ -34,7 +34,7 @@ def initialize(directory, max_size: nil, hasher: Refile::RandomHasher.new)
       # @return [Refile::File]      The uploaded file
       verify_uploadable def upload(uploadable)
         id = @hasher.hash(uploadable)
-        IO.copy_stream(uploadable, path(id))
+        ::File.open(path(id), "wb") { |dest| IO.copy_stream(uploadable, dest) }
```
scarfacedeb: I had to use {} block here, because jruby was raising: `syntax error, unexpected kDO_BLOCK` with multiline do..end block, and rubocop disallowed multiline {} blocks.

scarfacedeb:
I fixed a bug with a missing flush/close call with destination file in `def upload` by wrapping it in a block.
I was unable to run capybara-dependent tests in my environment (Is running `bundle exec rake`  enough?), but the rest of the tests are green, and travis.ci is passing.

lib/refile/backend/file_system.rb
jnicklas: We found a solution like this earlier through trial and error, but the problem is that this is a performance regression. When given a path, IO.copy_stream will do a straight file copy via syscalls, which is much faster than buffering the file through Ruby. I'd like to preserve that behaviour if at all possible.

scarfacedeb: I see. Then it depends on jruby fixes for this bug.
I'll use `Encoding.default_internal = nil` reset in my rails app as a temporarily solution for now.

PR #288, opened 2015-07-04
Remove unnecessary instruction

santib:
It isn't needed to add the `aws-sdk` gem since it is a dependency of the `refile-s3` gem.
`https://github.com/refile/refile-s3/blob/master/refile-s3.gemspec#L21`


janko-m:
Thanks @santib.

Issue #280, opened 2015-06-19
Image processing broken with GraphicsMagick


dburt:
Refile fails to process images when MiniMagick's underlying graphics processor is GraphicsMagick (1.3.18) rather than ImageMagick. This is because GraphicsMagick's mogrify command doesn't support the gravity option.

Related issue report on MiniMagick: https://github.com/minimagick/minimagick/issues/264

    $ gm version | head -1
    GraphicsMagick 1.3.18 2013-03-10 Q8 http://www.GraphicsMagick.org/
    $ rails c production
    Loading production environment (Rails 4.2.2)
    irb(main):001:0> MiniMagick.cli
    => :graphicsmagick

Stack trace for an image URL:

     INFO -- : Started GET "/attachments/store/fill/150/150/45c6a0a219a992365a105cead7aacbc145318bee51259dc8167e6ee6107a/photo" for 203.217.28.47 at 2015-06-19 
    ERROR -- : Refile::App: Error -> undefined method `gravity' for #<MiniMagick::Tool::Mogrify:0x00000003a2b8b0>
    ERROR -- : Refile::App: /var/www/wealthindex.co/shared/bundle/ruby/2.2.0/gems/refile-mini_magick-0.1.0/lib/refile/mini_magick.rb:68:in `block in fill'
    ERROR -- : Refile::App: /var/www/wealthindex.co/shared/bundle/ruby/2.2.0/gems/mini_magick-4.2.7/lib/mini_magick/image.rb:502:in `block in mogrify'
    ERROR -- : Refile::App: /var/www/wealthindex.co/shared/bundle/ruby/2.2.0/gems/mini_magick-4.2.7/lib/mini_magick/tool.rb:52:in `new'
    ERROR -- : Refile::App: /var/www/wealthindex.co/shared/bundle/ruby/2.2.0/gems/mini_magick-4.2.7/lib/mini_magick/image.rb:495:in `mogrify'
    ERROR -- : Refile::App: /var/www/wealthindex.co/shared/bundle/ruby/2.2.0/gems/mini_magick-4.2.7/lib/mini_magick/image.rb:372:in `combine_options'
    ERROR -- : Refile::App: /var/www/wealthindex.co/shared/bundle/ruby/2.2.0/gems/refile-mini_magick-0.1.0/lib/refile/mini_magick.rb:66:in `fill'
[...]

PR #290, opened 2015-07-06
Add a tiny note about GraphicsMagick compatibility

dburt:
GraphicsMagick can now be used instead of ImageMagick (See #280).


jnicklas:
Thanks!

PR #292, opened 2015-07-07
Prevent AJAX caching errors

goddamnhippie:
When using the multiple files feature on Safari (8.0.7 / OS X 10.10.4), sometimes the response to the presign url is cached resulting in the same upload id for multiple files, which in turn results in multiple files overwriting the same cached file.
This change fixes the issue by adding a querystring parameter to the presign url consisting of a timestamp and the file index, therefore avoiding duplicate urls.

### DIFF
--- a/app/assets/javascripts/refile.js
+++ b/app/assets/javascripts/refile.js
@@ -63,6 +63,7 @@
         if(input.getAttribute("data-presigned")) {
           dispatchEvent(input, "presign:start");
           var presignXhr = new XMLHttpRequest();
+          var presignUrl = url + "?t=" + Date.now() + "." + index;
           presignXhr.addEventListener("load", function() {
             dispatchEvent(input, "presign:complete");
             if(isSuccess(presignXhr)) {
@@ -77,7 +78,7 @@
               xhr.complete = true;
             };
           });
-          presignXhr.open("GET", url, true);
+          presignXhr.open("GET", presignUrl, true);
           presignXhr.send();
         } else {
           xhr.open("POST", url, true);


abitdodgy:
I can confirm that this happened to me today. I got a bunch of `Aws::S3::Errors::NoSuchKey`. Looking at the logs, I noticed that multiple files are being sent up with the same key. I've redacted most of the file keys but take my word for it that they are identical for both files.

```
{\"id\":\"[redacted] 76457b0c\",\"filename\":\"Lista Mestra de Documentos.pdf\",\"content_type\":\"application/pdf\",\"size\":299041},
{\"id\":\"[redacted] 76457b0c\",\"filename\":\"Controle de Registros.xlsx\",\"content_type\":\"application/vnd.openxmlformats-officedocument.spreadsheetml.sheet\",\"size\":28140}]"]}
````

**Comment by ivobenedito** [comment]
I'm also having this issue. 
Any idea when this PR will be merged?

Also, just as an opinion, shouldn't the presign calls go directly to the Refile Host instead of the CDN server?

**Comment by goddamnhippie** [comment]
@ivobenedito they have to go to S3 directly, that way you don’t use your app server’s resources. Heroku will even close the connection after 30 seconds so you can’t even upload medium-sized files on slow connections.

**Comment by ivobenedito** [comment]
@goddamnhippie The presign calls always need to go to the App Server. It's just an authorization call to allow the following direct S3 upload.

**Comment by goddamnhippie** [comment]
Oh, I thought you meant the upload call, the presign should go to the refile host.

**Comment by ivobenedito** [comment]
I've tried your solution and doesn't seem to be working. 
Doesn't Cloudfront ignore the query string?

**Comment by ivobenedito** [comment]
There is a temporary solution for this which doesn't need any change to the codebase.

By passing the **host option** on the **attachment_field** helper, the **data-url** gets built pointing to the Rails App host instead of the CDN.

```Ruby
<%= f.attachment_field :photos_files, host: request.base_url, multiple: true, direct: true, presigned: true %>
```

abitdodgy:
@ivobenedito how are you able to get the CDN to work? Have you seen issue #301?

goddamnhippie:
@ivobenedito if you copy this file https://github.com/goddamnhippie/refile/blob/master/app/assets/javascripts/refile.js to `/app/assets/javascripts/vendor/refile.js` and require it using `//= require /vendor/refile` on `/app/assets/javascripts/application.js` it should work, that’s all I did in my case.

This PR only solves the duplicate presign data issue.

ivobenedito:
@abitdodgy I'm using the **:host** option on the attachment_field helper to point the form presign calls to the Rails application instead of the CDN. 
That way, we don't have to fight caching :). The presign calls just shouldn't go to the CDN.

ivobenedito:
@goddamnhippie Yeah I know. You don't even need to require refile with the vendor prefix. It's just `//= require refile`.

But, in my case Cloudfront just ignores the query string timestamp your code is adding, so it still returns cached responses. I just changed the code to go directly to the rails app.

goddamnhippie:
@ivobenedito the presign calls never go to the CDN, they go to the sinatra app mounted on the rails app:
https://github.com/refile/refile/blob/master/app/assets/javascripts/refile.js#L32
https://github.com/refile/refile/blob/master/lib/refile/rails/attachment_helper.rb#L78
https://github.com/refile/refile/blob/master/lib/refile/app.rb#L72

ivobenedito:
@goddamnhippie https://github.com/refile/refile/blob/master/lib/refile/rails/attachment_helper.rb#L69

abitdodgy:
@ivobenedito you can setup your CDN to forward the query string, but if what your saying is true, then this is brilliant news. I've been fighting to get the CDN to work for a couple of days now.

@goddamnhippie but, in my case, with my CDN setup, I had a problem where they're goint to the CDN. See issue #301

ivobenedito:
@abitdodgy If CloudFront doesn't forward them by default, it should be for a good reason :P. There's no need to change CDN configuration if you can fix it the "right way".

goddamnhippie:
This patch solves a browser issue, the request is never getting to any server because Safari will return a cached response, therefore returning a duplicate key. In any case I am sure the presign request should never go to the CDN. Only the following upload request should. I’m not sure if that is what you guys mean.

ivobenedito:
@goddamnhippie Yes, that's what we mean and what the change I've pointed does.
Also, at least in my setup, it's not the browser caching the calls, it's the CDN. :v:

goddamnhippie:
Another note on the CloudFront querystring forwarding, I had to activate that because before my files were downloading with the wrong name (they were using the refile id), but I don’t know if that is always the case.

abitdodgy:
@ivobenedito so, if I understand you correctly:

1. You are successfully using a CDN
2. You don't have a problem with uploading multiple files, including files that have spaces in their name

And you managed this by adding `host: request.base_url` to the helper? Is there any thing special about your CloudFront setup?

@goddamnhippie I think something is wrong here, though. When I had the caching issue (duplicate keys, see first comment I made on this pull request) I wasn't using the CDN. This needs more investigating...

ivobenedito:
@abitdodgy That's right. I'm using Cloudfront with it's default settings and uploading ~ 100 images per day with the multiple option without any issue. Had to tweek CDN origin headers forwarding for serving static webfonts (CORS policies, but that is a different thing).

Meanwhile, not sure about files with spaces in their name ... never had that issue (yet).

jnicklas:
I succeeded in replicating the issue and verified that this PR fixes it. Thanks @goddamnhippie!

PR #317, opened 2015-08-02
add fallback option to attachment_url

musaffa:
Say one wants to use HTML5's srcset for [responsive image](http://demosthenes.info/blog/944/Responsive-Images-For-Retina-Using-HTML5s-srcset) display. To do that, multiple urls needs to be generated for different devices. `attachment_url` is perfect for this job except that it doesn't have any fallback option as is available in `attachment_image_tag`. This PR adds that option.

The fallback functionality could have been delegated from `attachment_image_tag` to `attachment_url` like this:
```ruby
def attachment_image_tag(record, name, *args, fallback: nil, format: nil, host: Refile.host, **options)
  classes = ["attachment", (record.class.model_name.singular if record), name, *options[:class]]
  image_tag(attachment_url(record, name, *args, fallback: fallback, format: format, host: host), options.merge(class: classes))
end
```
But the current method adds 'fallback' class to the fallback image_tag which I think is a good feature. So I didn't touch `attachment_image_tag`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,10 +369,11 @@ There's also a helper for generating image tags:
 <%= attachment_image_tag(@user, :profile_image, :fill, 300, 300) %>
 ```
 
-With this helper you can specify an image which is used as a fallback in case
+With this helper you can specify an image/asset which is used as a fallback in case
 no file has been uploaded:
 
 ``` erb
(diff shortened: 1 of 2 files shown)


jnicklas:
Sorry for the delay, just had a look at this and it looks good. Tests would have been nice, but since we don't have tests for the helpers anyway this'll have to do.

PR #324, opened 2015-08-17
Added support for receiving parsed json instead of just raw json strings

jabbawookiees:
We have some javascript that sends a POST request to Rails containing some metadata about a file pre-uploaded to S3 using the refile-s3 gem. The sent data follows the expected JSON format for Refile, but a gem we're using (JSONAPI::Resources) automatically parses that JSON.

The existing Refile code would assume that the data would be a Refile::File. This PR changes that behavior so that it can be smart about pre-parsed JSON.

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -71,7 +71,7 @@ module Refile
     end
 
     def set(value)
-      if value.is_a?(String)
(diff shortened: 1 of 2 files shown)


jnicklas:
Merged, thanks!

Issue #309, opened 2015-07-23
Handling soft deletes


abitdodgy:
We're using a soft deletes module that overrides ActiveRecord's `destroy` method. Instead of deleting the record, a `deleted_at` column is updated with a timestamp. Refile still calls `delete!` internally leaving the soft deleted file without an attachment. This can be a nasty surprise for those unaware of this behaviour.

See [line 44 in `lib/refile/attachment/active_record.rb`][1].

It's easy to solve this with a custom method, but it's worth leaving this issue out here for others. Maybe this should be configurable?

[1]: https://github.com/refile/refile/blob/20ecdb3d760b8eed2d5f4bddf6dd8c673063bf2c/lib/refile/attachment/active_record.rb#L44

janko-m:
I was thinking of the exact same feature recently, I need it when saving serialized copies of a record (snapshot), that in case the original gets deleted, the deserialized snapshot still shows the image.

I think the best place for this configuration is the [attachment definition](https://github.com/refile/refile/blob/master/lib/refile/attachment_definition.rb), just to add one more keyword argument.

jnicklas:
I was totally sure we had an option to switch off deletion of files, but it seems there isn't one. Weird.

shanemcd:
Does anyone have any recommendations for a workaround until this is resolved? I have tried `skip_callback :destroy, :after` with no luck.

shanemcd:
Update. I was able to hack around this by doing:

```ruby
attachment :file
define_callbacks :destroy
```

abitdodgy:
I have a pull request for this. https://github.com/refile/refile/pull/333

I don't expect it to be merged since it might be low quality, but I would welcome feedback from those involved in this issue.

abitdodgy:
Closing in favour of #362

tyleramos:
Looks like this issue was closed over a year ago, but this has not made it into a release yet.  The latest gem version 0.6.2 does not include this fix.  Any plans to release a new version of the gem?

abitdodgy:
@tyleramos my pull request was merged here: https://github.com/refile/refile/pull/362

tyleramos:
@abitdodgy Thanks for the quick response!  I noticed that it was merged into master but this has not been released as a ruby gem yet.  https://rubygems.org/gems/refile shows the latest version 0.6.2 was created in October but this was merged in November.  Any plans to release a new version of the ruby gem?

abitdodgy:
[...]

PR #333, opened 2015-08-24
Adds soft_delete configuration option

abitdodgy:
@jnicklas I don't expect this to get merged, but I would like your feedback so I can improve the feature. I added a config option to `refile.rb` called `soft_delete`; it defaults to `false`. I check for this option before running line https://github.com/refile/refile/blob/master/lib/refile/attachment/active_record.rb#L44, when then `after_delete` callback executes. See issue #309 for reference.

Some feedback would be nice...

1. It feels wrong to add `soft_delete` to the config when it only works with ActiveRecord. It seems like `soft_delete` should work regardless of the adapter. The only other place I can think of implementing this is https://github.com/refile/refile/blob/master/lib/refile/attacher.rb#L134. Perhaps I'm wrong.
2. Should there be more tests? For example, that `soft_delete` defaults to `false` and that it can be set? Where would we add those? `refile_spec.rb`?
3. Is the one test in `spec/refile/attachment/active_record_spec.rb` sufficient for this feature?
4. Is `soft_delete` a good name?

### DIFF
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -94,6 +94,13 @@ module Refile
     # @return [String]
     attr_accessor :secret_key
 
+    # Should we keep files after a record is deleted?
+    #
+    # The default is false.
+    #
+    # @return [Boolean]
+    attr_accessor :soft_delete
+
     # A global registry of backends.
     #
     # @return [Hash{String => Backend}]
@@ -369,4 +376,5 @@ Refile.configure do |config|
   config.automount = true
   config.content_max_age = 60 * 60 * 24 * 365
   config.types[:image] = Refile::Type.new(:image, content_type: %w[image/jpeg image/gif image/png])
+  config.soft_delete = false
 end
--- a/lib/refile/attachment/active_record.rb
+++ b/lib/refile/attachment/active_record.rb
@@ -42,7 +42,9 @@ module Refile
         end
 
         after_destroy do
-          send(attacher).delete!
+          unless Refile.soft_delete
+            send(attacher).delete!
+          end
         end
       end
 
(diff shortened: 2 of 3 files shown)


jnicklas:
1. I would rather have this as a config option to the `attachment` method, rather than a global config.
2. Yes, a few more tests would be good, I think `spec/refile/attachment/active_record_spec.rb` is the right place.
4. I've wanted this kind of option for a while (in fact I though it already existed and was surprised to find out it didn't), but I think it's more general than soft deletion. Someone might want to keep files around just in case, without implementing a soft deletion strategy. I think this is quite common. Some options could be:

``` ruby
attachment :foo, old_files: :keep # explicit but a bit wordy
attachment :foo, prune: true # difficult to understand, assumes the user knows what prune means
attachment :foo, destroy: true # destroy what exactly?
attachment :foo, dependent: :delete # Rails-alike, but maybe a bit confusing?
```

Not sure I really super like any of these.

abitdodgy:
@jnicklas thanks for the feedback. I like `attachment :foo, dependent: :delete`, but I think it can be confusing.  IMO `old_files: :keep` is too verbose. How do you feel about:

```ruby
attachment :foo, keep_file: true    # seems explicit enough
attachment :foo, file: :destroy
# or, depending on which default behaviour we want... ?
attachment :foo, file: :destroy
```
Also, should `file` not be singular? We're dealing with one.

jnicklas:
After thinking about this some more, I think that maybe I like `destroy: true|false` the best. What do you think?

abitdodgy:
I think that sounds good! I'll start a new branch, and close this one once I make a new pull request.

abitdodgy:
Closing in favour of #362

Issue #343, opened 2015-09-06
Auth token should be checked before retrieving file from backend (and process it)


fabn:
I like this library and its approach and I'm planning to replace carrierwave with this, however imho there is a severe issue in its token validity check.

Token authentication is very good way to prevent unauthorized and potentially resource expensive calls. Imagine a request that resize an image at 100000x100000, this will probably kill a lot of servers (especially on resource constrained environments like heroku).

For this reason you should really move token validation in a [before filter](http://www.sinatrarb.com/intro.html#Filters) instead of [checking it](https://github.com/refile/refile/blob/master/lib/refile/app.rb#L122) after file has been [retrieved or processed](https://github.com/refile/refile/blob/master/lib/refile/app.rb#L35:L58)

fabn:
For the record, I just tried to change a fill url using `fill/100000/100000/` in my development machine and refile created a 80GB file only to return a forbidden response.

PR #345, opened 2015-09-07
Fix for #343

fabn:
Here's a quick fix with spec for the issue I opened in #343

Let me know if it's ok.


jnicklas:
Sounds good! Thank you!

PR #353, opened 2015-09-10
Update README.md

callmeed:
Fixed protocol-relative note to mention .cdn_host (instead of the old .host) option.


jnicklas:
Thanks!

PR #356, opened 2015-09-11
updates quickstart to include all necessary migration fields

superacidjax:
The current quickstart guide neglects to include necessary fields. If the quickstart were followed precisely, it results in an error.


rilian:
:+1:

Issue #355, opened 2015-09-10
Presence validation for #{name}_id interferes with attacher.store!


monkbroc:
I wanted to make sure a file is always attached to the model so I used this validation:

```ruby
class Attachment < ActiveRecord::Base
  attachment :file
  validates_presence_of :file_id, :file_filename, :file_size, :file_content_type
end
```

When I save an Attachment, it fails because the file_id is empty, but the other attributes are filled in. I figured out that the `attacher.store!` method called in a `before_save` hook is supposed to set the `#{name}_id`.

Removing the `file_id` presence validation works around the issue.

How would you make sure that a file is always present?

abitdodgy:
I would deal with this via a database constraint. There should be no circumstance where the file id is missing because of user input, and errors in user input is what validations are meant to catch. In this case I would want a failure, not a validation error. If the `file_id` is missing it's likely because of a bug, and not something that the user has done, so giving the user feedback about the error doesn't really gain you anything.

jnicklas:
You should  just be able to:

``` ruby
class Attachment < ActiveRecord::Base
  attachment :file
  validates_presence_of :file
end
```

abitdodgy:
I take that back, actually. There obviously are circumstances when this could happen. My thinking was so ingrained in my current use of Refile that I've forgotten about obvious scenarios.

monkbroc:
`validates_presence_of :file` leads to some quirks. I'll close this and we can continue the discussion in PR #359.

PR #359, opened 2015-09-11
Better support for validating presence of attachment

monkbroc:
As discussed in #355, to validates the presence of an attachment, you should be able to do
```ruby
class Attachment < ActiveRecord::Base
  attachment :file
  validates_presence_of :file
end
```

Setting `attachment.file = nil` raises the exception `undefined method 'size' for nil:NilClass`. This PR fixes this by making `attachment.file = nil` mean the same thing as `attachment.remove_file = true`.

Setting `attachment.file = nil` or `attachment.remove_file = true` will still return the old file in `attachment.file` so the validation passes even though the file will be removed after save. This PR returns `nil` for `attachment.file` if `attacher.remove? == true`.

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -74,9 +74,9 @@ module Refile
 
     def set(value)
       case value
-      when nil then self.remove = true
-      when String, Hash then retrieve!(value)
-      else cache!(value)
+        when nil then self.remove = true
+        when String, Hash then retrieve!(value)
+        else cache!(value)
       end
     end
 
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -63,7 +63,9 @@ module Refile
     end
 
     def get
-      if cache_id
+      if remove?
+        nil
+      elsif cache_id
         cache.get(cache_id)
       elsif id
         store.get(id)
@@ -71,10 +73,10 @@ module Refile
     end
 
     def set(value)
-      if value.is_a?(String) or value.is_a?(Hash)
-        retrieve!(value)
-      else
-        cache!(value)
+      case value
+      when nil then self.remove = true
+      when String, Hash then retrieve!(value)
+      else cache!(value)
       end
     end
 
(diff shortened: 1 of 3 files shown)


jackturnbull:
Can't comment on the code within the PR as I haven't had chance to properly look at it but functionally this gets a +1 from me.

The current use of `validates_presence_of` directly on the attachment itself is incompatible with the shoulda-matcher [`validates_presence_of`](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb).

monkbroc:
> The current use of validates_presence_of directly on the attachment itself is incompatible with the shoulda-matcher validates_presence_of.

This is what prompted me to make this PR.

jackturnbull:
@monkbroc Is there any scenarios where user input would cause the uploaded file to be `nil`, rather than an empty string? As far as I can tell, if I have it as a required param, then this should always come from the controller as `''` rather than `nil` so I should be covered.

I think that there are inconsistencies that need to be resolved (hence the PR) but I'm debating how to guard against these in my app, or whether it's worth testing against these edge cases. How did you decide to approach this? Assuming you're not using your fork that is.

monkbroc:
If the parameter is present in the request, it cannot be nil as you say. If it's not present, the `file=` method won't be called in `assign_attributes` so that's ok too.

I'm using my fork of course :laughing: 

Honestly this is more to make shoulda-matcher work well. It's not a situation that normally shows up so I feel this PR is low risk.

jackturnbull:
> Honestly this is more to make shoulda-matcher work well. It's not a situation that normally shows up so I feel this PR is low risk.

Valuable to me also!

Cheers for the response, looks like I'll wrap it in a pending for now.

jnicklas:
Looks good to me! Merging :) Sorry for the delay!

monkbroc:
No worries. Thanks for taking time from your day to work on some open source :+1:

Issue #334, opened 2015-08-24
Presenters and attachment_image_tag


darkleaf:
I use presenters based on Delegator class.
My object has structure like this: (UserPresenter ( AccessPresenter ( User ))).
I pass ( AccessPresenter ( User )) to the attachment_image_tag helper.
This heper use [record.class.model_name.singular](https://github.com/refile/refile/blob/master/lib/refile/rails/attachment_helper.rb#L41).
But my presenter class haven't model_name method.

In rails 4.2 we can write like this: `record.model_name.singular` because [instance delegate model name to it class](https://github.com/rails/rails/blob/7847a19f476fb9bee287681586d872ea43785e53/activemodel/lib/active_model/naming.rb#L219)

How I can help fix this?
I can made PR, that uses if and respond_to? method in this helper:

```ruby
classes = ["attachment", name, *options[:class]]
classes << record.model_name.singular if record.respond_to? :model_name
classes << record.class.model_name.singular if record.class.respond_to? :model_name
classes.uniq!
```

jnicklas:
I don't like that this silently behaves differently in different versions of Rails. Can't you just `extend ActiveModel::Naming` in your presenter?

darkleaf:
I think about it. This solution  can't help.

```ruby
class BasePresenter < Delegator
  attr_reader :model, :h
  alias_method :__getobj__, :model

  def initialize(model, view_context)
    @model = model
    @h = view_context
  end
end

class UserPresenter < BasePresenter
  def initialize(model, view_context)
    access_presenter = AccessPresenter.new model, view_context
    super access_presenter, view_context
  end

  def profile_image
    return unless model.profile_image
    h.attachment_image_tag(model, :profile_image, :fill, 200, 200)
  end
end
```
`model` is a `AccessPresenter` instance.
`AccessPresenter.model_name.singular` method  will return  'access_presenter'.
`model.model_name.singular` will return  'access_presenter' too, because instance delegate `model_name` method to it class(AccessPresenter). 

But `model.model_name.singular` must return 'user'.

```ruby
user = User.new
presenter = UserPresenter.new user, view_context
presenter.model_name.singular #=> 'user'

presenter.model #=>  AccessPresenter instance
presenter.model.model_name.singular #=> 'user'

presenter.model.model #=> User instance
presenter.model.model.model_name.singular #=> 'user'

class AccessPresenter
  extend ActiveModel::Naming
end

user = User.new
presenter = UserPresenter.new user, view_context

[...]

PR #360, opened 2015-09-13
add presenter support for attachment_image_tag helper

darkleaf:
I find solution for #334. It use `to_model` method and behaves in the same way in different versions of Rails.


jnicklas:
I don't like this. This adds the assumption that all recods respond to `to_model` and return themselves. This is true of ActiveRecord models, but that doesn't cover any other non-AR model objects someone might pass in, including objects from other ORMs, or possibly form objects or whatever else.

darkleaf:
`to_model` method of `ActiveModel::Conversion`.
`model_name` method of `ActiveModel::Naming`.
`model_name` used in `attachment_image_tag`
So model must use ActiveModel.

But if we wont use PORO objects as model,  `attachment_image_tag` helper should not be use `model_name` method.

sobrinho:
Closing due to the lack of activity.

PR #362, opened 2015-09-14
Adds `destroy: true | false` option to `attachment` method

abitdodgy:
@jnicklas As discussed in #333 this adds a `destroy: true | false` option. It defaults to true. I'm unsure if the documentation is in the right place. Any feedback is welcome!

### DIFF
--- a/lib/refile/attachment.rb
+++ b/lib/refile/attachment.rb
@@ -34,8 +34,9 @@ module Refile
     # @param [Symbol, nil] type                         The type of file that can be uploaded, see {Refile.types}
     # @param [String, Array<String>, nil] extension     Limit the uploaded file to the given extension or list of extensions
     # @param [String, Array<String>, nil] content_type  Limit the uploaded file to the given content type or list of content types
+    # @param [true, false] destroy                      Whether to remove stored file if its model is destroyed
     # @return [void]
-    def attachment(name, cache: :cache, store: :store, raise_errors: true, type: nil, extension: nil, content_type: nil)
+    def attachment(name, cache: :cache, store: :store, raise_errors: true, type: nil, extension: nil, content_type: nil, destroy: true)
       definition = AttachmentDefinition.new(name,
         cache: cache,
         store: store,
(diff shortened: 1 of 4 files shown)


abitdodgy:
@jnicklas not sure why Travis is failing here... :disappointed:

lib/refile/attachment.rb
```diff
@@ -34,8 +34,9 @@ module Attachment
     # @param [Symbol, nil] type                         The type of file that can be uploaded, see {Refile.types}
     # @param [String, Array<String>, nil] extension     Limit the uploaded file to the given extension or list of extensions
     # @param [String, Array<String>, nil] content_type  Limit the uploaded file to the given content type or list of content types
+    # @param [true, false] destroy                      Whether to remove stored file if its model is destroyed
     # @return [void]
-    def attachment(name, cache: :cache, store: :store, raise_errors: true, type: nil, extension: nil, content_type: nil)
+    def attachment(name, cache: :cache, store: :store, raise_errors: true, type: nil, extension: nil, content_type: nil, destroy: true)
```
jnicklas: Seems like Rubucop complains about the unused argument here. I suppose if we don't accept it then we have to explicitly pass the right options to `super`, though that might still be the better solution?

abitdodgy: I can change that. I think it's best to be explicit. I'll move the documentation up to `lib/refile/attachment/active_record.rb`. So something like this?

```ruby
def attachment(name, raise_errors: false, destroy: true, **options)
  super(name, raise_errors: false, **options)
```

abitdodgy:
@jnicklas do you think can we merge this?

mmplisskin:
I think this would be a great option for those of us doing backups so we can just store file_id and backup images in s3.

trentclowater:
Any idea if this will be merged soon?

abitdodgy:
@jnicklas sorry to be pest, but is there anything else I can here for this to be merged? I fixed the rubocop issue a while ago, and a few people are asking for this feature!

jnicklas:
Merged. Hopefully I'll have some time to go through other long standing issues in the next couple of days.

abitdodgy:
@jnicklas thanks a bunch for taking care of this! I appreciate it! :tada: :+1:

PR #373, opened 2015-09-26
add missing parameters to `attachment_image_tag` doc [ci skip]

y-yagi:


lib/refile/rails/attachment_helper.rb
```diff
@@ -38,8 +38,13 @@ def attachment_url(record, name, *args, fallback: nil, **opts)
...
-    # @param [String] fallback                   The path to an image asset to be used as a fallback
-    # @param [Hash] options                      Additional options for the image tag
+    # @param [Refile::Attachment] record    Instance of a class which has an attached file
+    # @param [Symbol] name                  The name of the attachment column
+    # @param [String, nil] fallback         The path to an image asset to be used as a fallback
+    # @param [String, nil] host             Override the host
+    # @param [String, nil] prefix           Adds a prefix to the URL if the application is not mounted at root
+    # @param [String, nil] format           A file extension to be appended to the URL
+    # @param [Hash] options                 Additional options for the image tag
     # @see #attachment_url
```
jnicklas: The idea was that the link to `attachment_url` would provide all the other options, so we don't have to repeat them. I'm not super against repeating them though, maybe it's not quite clear how it is now?

y-yagi: Thank you for explanation. I understood a meaning. 

Since the `fallback` option has been described in the methods of both `attachment_image_tag` and `attachment_url`, even than necessary other options and I was mistaken. 
I thought that there should be explanation to perform reference of a link about other options.

PR #384, opened 2015-10-20
Recommend using https for the cdn_host

sorentwo:
As described in the most recent update to http://www.paulirish.com/2010/the-protocol-relative-url/, it is recommended to always use HTTPS.


jnicklas:
Thanks!

PR #389, opened 2015-11-01
Add a reference to gridfs backend in the readme.

Titinux:
I've released a refile backend for Gridfs. Is it possible to add a reference to it in the readme ? Thanks.


jnicklas:
Awesome :) Thanks!

PR #394, opened 2015-11-23
Make Refile::FileDouble available to gem users

monkbroc:
When I write specs for my own models using Refile, I want an easy way to provide files. The easiest way is to use the Refile::FileDouble class, but currently it's hidden inside the spec helper.

This PR puts FileDouble in the lib directory so it is available to gem users.

If you like the idea I can add a small section to the README about testing your own classes.


jnicklas:
Nice, thank you! A README entry wouldn't be bad.

PR #396, opened 2015-11-24
Add testing example with FileDouble to README

monkbroc:
I added an example of how to test a model using Refile attachments using the `Refile::FileDouble` helper instead of real files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -780,6 +780,42 @@ needs:
 - Choose whatever number of days you're comfortable with, I chose "1"
 - Click "Review" and finally "Create and activate Rule"
 
+## Testing
+
+When testing your own classes that use Refile, you can use `Refile::FileDouble` objects instead of real files.
+
+```ruby
+# app/models/post.rb
+class Post < ActiveRecord::Base
+  attachment :image, type: :image
+end
+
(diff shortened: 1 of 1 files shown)


abitdodgy:
@monkbroc thanks for the three pull requests you've made. They've made a big difference to me here. (and thanks to @jnicklas for merging them). @monkbroc your testing example was very helpful. Do you mind sharing with me how you go about testing the controller? I'm doing something like this:

```ruby
test "foo" do
  file = Refile::Backend::FileSystem.new("tmp/uploads/cache").upload StringIO.new('hello')
  file_attributes = {
    id: file.id,
    size: file.size,
    filename: 'hello.txt',
    file_content_type: 'text'
  }.to_json

  post :foo, {
    file: file_attributes
  }

  # assertions
end
```

As you can see I'm building the json manually and passing it to the params hash. Have you found a better way?

monkbroc:
@abitdodgy I haven't done any controller tests for Refile. I use Capybara for integration tests and upload real files.

abitdodgy:
@monkbroc thanks! It probably makes more sense to do that!

PR #398, opened 2015-12-02
Always return 304 Not Modified when the browser asks for updated file

monkbroc:
Since Refile attachments are immutable I think it makes sense to always return 304 Not Modified when a browser asks if a file has changed since it was last downloaded.

The practical aspect is that resized images are loaded only once, reducing the CPU load of resizing. This is nice in development. I can't speak to the performance increase in production with a CDN but it should help there too when several CDN edge servers ask for the same resource.

If you think this is a good idea I'll write tests for this change.

### DIFF
--- a/lib/refile/app.rb
+++ b/lib/refile/app.rb
@@ -35,6 +35,7 @@ module Refile
     # This will match all token authenticated requests
     before "/:token/:backend/*" do
       halt 403 unless verified?
+      halt 304 if has_if_modified_since?
     end
 
     get "/:token/:backend/:id/:filename" do
@@ -90,6 +91,11 @@ module Refile
       "forbidden"
     end
 
+    error 304 do
+      content_type :text
+      "not modified"
+    end
+
     error Refile::InvalidFile do
       status 400
       "Upload failure error"
@@ -117,6 +123,10 @@ module Refile
 
     def upload_allowed?
       Refile.allow_uploads_to == :all or Refile.allow_uploads_to.include?(params[:backend])
+    end
+
+    def has_if_modified_since?
+      !request.env["HTTP_IF_MODIFIED_SINCE"].blank?
     end
 
     def logger


jnicklas:
I'm a bit wary of how this affects CDNs, do you have any experience with that?

monkbroc:
Not yet. Let me turn on the CDN host on my site and report back what I find.

In principle though, do you think this is a good idea? A specific attachment_url will always point to exactly one version of a file, right?

jnicklas:
I think that this is possibly unnecessary, I'm not sure if we even get any if-modified-since headers.

jnicklas:
If it is a legitimate improvement I'm all for it though.

monkbroc:
With the CDN configured I can't see any change in behavior. As you noted returning 304 is unnecessary in this case: the CDN caches the file and each edge server that needs it will request it only once. It never sends if-modified-since headers.

The real use for me is in development: when I load a profile index page full of avatars, each takes ~1 second to resize each one, so the full page load is ~30 seconds.

With the 304 Not Modified change each image is loaded only once unless I do a hard reload. Subsequent profile page loads are therefore very fast.

Is there a better way to have fast resized image load in development without sending 304 Not Modified?

jnicklas:
I'm really surprised this would have any effect in development either, since we already have far-future cache expiry headers. Are you sure you're on the latest version of Refile otherwise? I think I'm going to close this for now.

monkbroc:
Thanks for looking back at this issue Jonas. I still think the issue is there.

I agree that refile sends the correct cache expiry headers.

The issue is that refile doesn't check those headers when a browser makes the request so it always processes images and sends the entire body of the file even when the browser indicates it only wants a file if it has changed. Since refile attachments are immutable, it can never have changed if a browser indicates it has a cached copy.

I run into this in development when reloading a page that has refile `attachment_image_tag` with F5. The browser reloads all attachment images even with the far-future cache expiry headers.

I made a repository that illustrates this with the latest version of refile. Please check it out when you have time.
https://github.com/monkbroc/refile-not-modified

PR #402, opened 2015-12-15
Add option to save processed files in local public folder

aglushkov:
I add an option `Refile.public_path` so everybody who is not gang ho about CDN can preserve processed files in public folder. 
With this change refile would not process the same files over and over with each page reloading.
If you will not set this option than all should work like before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,6 +342,18 @@ elloh
 Refile calls `call` on the processor and passes in the retrieved file, as well
 as all additional arguments sent through the URL.
 
+### Saving processed files in public folder
+
+If using CDN is not an option then you probably want to save processed files
+in public folder.
+You should only define `public_path` option like this:
+
+``` ruby
+Refile.public_path = '/your/public/folder'
+# or
+Refile.public_path = Rails.public_path if Rails.env.development?
+```
+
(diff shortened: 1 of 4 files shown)


jnicklas:
Even if you wanted to do this conceptually, this is absolutely *not* the right approach. Adding this in the application is completely wrong. You should specify this in the backend. In fact, it should be as easy as setting up a file storage backend with a root directory which is in your public path.

That being said, it's not a strategy I would recommend. At least you should set up something to distribute the files among app servers, otherwise you have effectively locked yourself to never scaling horizontically, which is bad, imo.

jnicklas:
I am closing this PR. Thank you for your effort, but this is just not a good fit to our goals.

aglushkov:
@jnicklas, Refile is best, and thank you for such detailed explanation of your point.

Visiting the same image-heavy pages on **localhost** makes me and my PC suffer. So I monkey-patched refile to preserve processed files if public folder. No more suffer. Recently I have known that my friend have the same issue. So I decided to make pull request to show how I solve it.
Hope this helps to somebody else.

PS: File system backend is responsive for saving original files only, not processed. So it can't help.

Issue #270, opened 2015-06-09
Clash with attachment_url method


abitdodgy:
I'm justing putting this out there for discussion. We have an Attachment model, and some of its controller routes are clashing with `attachment_url` that Refile uses. We'll either add a namespace, or nest attachment routes under a parent model. Just food for thought, but I wonder how much of an impact this has on older applications have have `attachments_controller` and so forth. I our case we're early in the building stages of our app, so the cost of making this change is rather cheap.

jnicklas:
We could provide an alias called something like `refile_attachment_url` or something, what do you think about that?

abitdodgy:
@jnicklas IMO that would be great to have. Would you like a pull request? This might be a bit out of my depth, to be honest, but we all have to start somewhere, so I don't mind giving it a bash when I have some time. What do you think?

jnicklas:
Yeah, give it a shot! :)

monkbroc:
I ran into this as well. I made an `Attachment` model with a polymorphic association to a couple other models in my app. I ended up renaming it to `FileAttachment` just to work around the name collision with the `AttachmentHelper#attachment_url` helper.

So this is going to keep happening to others.

I'm not sure what could be done to fix this since renaming `attachment_url` would break existing applications.

toothfairy:
I believe attachment_url is a very abstract and common method name to use. Perhaps ```refile_url``` might be replacement for it in some future. It directly references to it's nature and would help to avoid names collisions. @jnicklas what do you think about it?

abitdodgy:
@toothfairy one of the suggestions that @jnicklas already made is `refile_attachment_url`, which I think is a better suggestion. I think `refile_url` is too ambiguous.

toothfairy:
Not sure in what way it's ambiguous, @abitdodgy, but may be you're right. In my opinion ```__attachment__``` part in refile_attachment_url makes it way to explicit and long. In context of refile,  method named ```*_url``` can barely mean something else than attachment url. 

Also making alias won't solve the problem entirely (there still would be two methods with the save name in a different parts of application). My idea was to rename ```attachment_url``` to ```refile_url``` (or ```refile_attachment_url```)

abitdodgy:
@toothfairy `refile_url` could be mistaken for where you mount your refile app, for example. But why would there be more than one `refile_attachment_url`?

rilian:
I have "fixed" this issue in my project by explicitly calling `Refile.attachment_url` :)

[...]

PR #403, opened 2015-12-15
Add aliases to helper methods

toothfairy:
As it was discussed at #270 this adds alias for helper methods and alias to attachment_url from refile module to make it consistent.

I'm not sure if it would be tested since alias_method is language construction

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,6 +392,8 @@ no file has been uploaded:
 <%= attachment_image_tag(@user, :profile_image, :fill, 300, 300, fallback: "default.png") %>
 ```
 
+All helper methods are available with `refile_` prefix to avoid name clashes
+
 ## 5. JavaScript library
 
 Refile's JavaScript library is small but powerful.
--- a/lib/refile.rb
+++ b/lib/refile.rb
@@ -391,6 +391,8 @@ module Refile
       file_url(file, *args, host: host, prefix: prefix, filename: filename, format: format)
     end
 
+    alias_method :refile_attachment_url, :attachment_url
+
     # Receives an instance of a class which has used the
     # {Refile::Attachment#attachment} macro to generate an attachment column,
     # and the name of this column, and based on this generates a URL to a
--- a/lib/refile/rails/attachment_helper.rb
+++ b/lib/refile/rails/attachment_helper.rb
@@ -32,6 +32,8 @@ module Refile
       end
     end
 
+    alias_method :refile_attachment_url, :attachment_url
+
     # Generates an image tag for the given attachment, adding appropriate
     # classes and optionally falling back to the given fallback image if there
     # is no file attached.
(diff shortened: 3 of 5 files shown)


abitdodgy:
Correct me if I'm wrong, but doesn't this just add an alias to `attachment_url` instead of solving the actual problem?

As for testing, you would need to test that Refile responds to any new methods, aliased or otherwise.

toothfairy:
@abitdodgy as I understood form #270, you proposed to add alias (and you're correct, it doesn't sove the problem entirely). But is it better to deprecate current method names and to remove them later?

abitdodgy:
@toothfairy then my original proposal was wrong. I think I'll defer to @jnicklas on this one. Deprecating the existing methods sounds reasonable to me.

jnicklas:
In principle I am for this, and I am against deprecating the existing methods. These should be purely aliases available in the rare case of a naming conflict.

@toothfairy have you actually checked the implementation though? There are subtle differences between `alias` and `alias_method` that I can never remember which might affect the right course of action here.

toothfairy:
@jnicklas I've added some tests for alias methods.

Also talking about name conflicts, I'd agree with @abitdodgy, alias might be helpful but it doesn't solve problem entirely.

abitdodgy:
@jnicklas am I missing something? But as pointed out before, just aliasing the methods doesn't really solve the problem, does it?

jnicklas:
@abitdodgy not sure what you mean, this seems like a pretty good solution to me. As far as I understand the original problem, we were trying to solve name collisions between e.g. `attachment_url` from Refile and from a Rails route for a model called Attachment. If we simply provide `refile_attachment_url` as an alternative, that should solve the problem, should it not?

abitdodgy:
@jnicklas from what I can see in the PR `refile_attachment_url` is just an alias of `attachment_url`, which means `attachment_url` is still being used by Refile. Maybe I'm missing something?

toothfairy:
The initial idea was to remove `attachment_url` method at all but it's kind of a big change.

toothfairy:
@abitdodgy so adding alias should actually fix your problem. Engines are included first (and helpers from engines). If you have `resources :attachments` in your routes file (or something else which re-defines attachment_url) you're `attachment_url` will be re-defined (because method lookup, right?). But alias `refile_attachment_url` will perform lookup in lexical scope at first place and will refer to your original `attachment_url` method

abitdodgy:
@toothfairy then that's the information I was missing, thanks!

rilian:
I have also stumbled upon this issue, since i have `Attachment` models and I'm just using `Refile.attachment_url(...)` in my app views. Isn't it simpler than making an alias ?

Sorry for double posting, but it looks like a popular issue here https://github.com/refile/refile/issues/270

sobrinho:
Anyone can rebase it against master and see if it still works?

sobrinho:
Closing since #430 got merged, use `Refile.attachment_url(...)` in case of collision.

Thanks!

Issue #335, opened 2015-08-26
Not working for nested forms, missing index


jokklan:
Hi

I can't remove or update existing images on a nested model after first upload. The problem seems to be that there is missing a index on the html tag name attribute on the hidden field.

Example:
```erb
<%= form_for @user do |f| %>
  <%=  f.fields_for :images do |image_form| %>
    <%= image_form.attachment_field :file %>
  <% end %>
<% end %>
```
results in something like:
```html
<form id="edit_user_1" enctype="multipart/form-data" action="/users/1" method="post">
  <!-- This is where the index [1] is missing in the name attribute -->
  <input value="null" type="hidden" name="user[image_attributes][file]"> 

  <input name="user[image_attributes][1][file]" id="user_image_attributes_1_file">
</form>
```

I'm not sure how this can be solved, i looked https://github.com/refile/refile/blob/v0.5.5/lib/refile/rails/attachment_helper.rb but can't figure out why `hidden_field` and `file_field` behaves differently.

UPDATE:
To clarify, this problems is only for nested attributes through a `have_many` association.
```ruby
class User < ActiveRecord::Base
  has_many :images

  accepts_nested_attributes_for :images
end
```

simonwh:
I have this exact same problem. Any ideas on what we can do to fix it?

simonwh:
@jnicklas Do you have any ideas on how to go about solving this?

sidtri:
It seems fine for me to adding, updating records. Can you give some more content related to this issue.

```html
<div class="field">
    
     <input type="text" value="df" name="user[post_attributes][name]" id="user_post_attributes_name">
     <input value="{}" data-reference="8c74cccf68793046d2072efa23faeb6a" type="hidden" name="user[post_attributes][image]"><input data-reference="8c74cccf68793046d2072efa23faeb6a" type="file" name="user[post_attributes][image]" id="user_post_attributes_image">
<input type="hidden" value="1" name="user[post_attributes][id]" id="user_post_attributes_id">  </div>
```

Values are populating fine for me

jokklan:
@sidtri you only have óne related record (`has_one`/`belongs_to`), the problem is when you have many nested models.

jnicklas:
Is this issue solved by #406? Can we close this?

jokklan:
@jnicklas yes it is, thanks! I'll close it :)

PR #406, opened 2015-12-17
Pass :index option through to hidden field

mattbrictson:
Currently, if we call `attachment_field(:image, index: 1)`, Refile omits the `:index` option when rendering the hidden field. This means the hidden field and file field are rendered with different names, e.g.:

```
<input type=hidden name=object[image] ...>
<input type=file   name=object[1][image] ...>
```

This is a bug.

The hidden field and file field need to render with identical `name` attributes in order for file upload to work properly. The name is generated by Rails based on certain options, so we need to make sure that both the field and file receive the same options, at least those options that affect the name. `:index` is one of those options. 

This PR fixes this discrepancy by passing the `:index` option to the hidden field so that the output is correct:

```
<input type=hidden name=object[1][image] ...>
<input type=file   name=object[1][image] ...>
```

### DIFF
--- a/lib/refile/rails/attachment_helper.rb
+++ b/lib/refile/rails/attachment_helper.rb
@@ -84,13 +84,16 @@ module Refile
 
       options[:data][:reference] = SecureRandom.hex
 
-      hidden_field(object_name, method,
+      hidden_options = {
         multiple: options[:multiple],
         value: object.send("#{method}_data").try(:to_json),
         object: object,
         id: nil,
         data: { reference: options[:data][:reference] }
-      ) + file_field(object_name, method, options)
+      }
+      hidden_options.merge!(index: options[:index]) if options.key?(:index)
+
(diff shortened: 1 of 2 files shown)


jnicklas:
Nice :) Thank you! It seems like this is a solution to #335? 

I'm not sure if there's any sensible way of testing this, some kind of regression test would be super nice, but maybe not absolutely required for me to merge this.

mattbrictson:
@jnicklas I've updated the PR and added a test. There wasn't yet a test for `attachment_helper`, so I added one that covers the `:index` scenario. I hope that is the right place to put it.

jokklan:
Ahh couldn't figure this bug out! Awesome really needed that fix, thanks +1

simonwh:
:+1: Fantastic, also really needed this.

jnicklas:
Awesome :) Thanks!

PR #407, opened 2015-12-21
Remove metadata

### LINKED ISSUE #401: delete attachment
I have tried to add checkbox for the purpose to delete attached file.
I have  the following columns:
commented_id
commented_filename
commented_size
commented_content_type

actually only commented_id sets to NULL, other columns still maintain their values. Any suggestion?
thanks
Fabio

aglushkov:
This fixes #401, attachment metadata was not removed with attachment

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -128,12 +128,13 @@ module Refile
       if remove?
         delete!
         write(:id, nil, true)
+        remove_metadata
       elsif cache_id
         file = store.upload(get)
         delete!
         write(:id, file.id, true)
+        write_metadata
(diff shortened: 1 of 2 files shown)


jnicklas:
The spec should probably be in `spec/refile/attachment_spec`, I think? There doesn't seem to be anything ActiveRecord specific in this from what I can tell.

aglushkov:
Yes, a kind of redundant test. I will remove this from active_record_spec, and add one to attachment_spec

aglushkov:
Ready

jnicklas:
Awesome :) Thanks!

PR #423, opened 2016-01-19
Added Spanish, German, French and Japanese translations.

kobaltz:


jnicklas:
I do not want to add translations to the gem, it adds too much of a maintenance burden. Whenever we change something which affects these, we need to find someone proficient in all the languages we have added, that's unsustainable. I'd be fine with gathering these all up and linking them or something, but I don't like adding these to the gem itself. Other Ruby gems which have translation files, like Devise, seem to follow the same policy.

PR #430, opened 2016-01-30
improve Readme and fix some typos

rilian:
1) added notes for case when project has `attachment_*` routes
2) added example of overriding default store folder
3) fixed some typos


sobrinho:
Thanks! ❤️

rilian:
hi, that was quick 🐎

sobrinho:
Better late than never 😹 😹 😹

PR #433, opened 2016-02-07
Use different escape method for filename

toothfairy:
This change will make it so escape doesn't replace " " to "+" and downloaded files will have more readable names. (ie `test+example.txt` vs `text example.txt`)


toothfairy:
https://github.com/rack/rack/blob/master/lib/rack/utils.rb#L35

sobrinho:
Thanks! ❤️

PR #436, opened 2016-02-15
Fix gem releasedate

seapy:
Version 0.6.2 release date is 2015-10-29 according to [rubygems](https://rubygems.org/gems/refile)


sobrinho:
Thanks! <3

Issue #429, opened 2016-01-28
access uploaded images url from javascript


tonklon:
since 0.6 we can no longer construct the attachment_url on the client side. I understand why it has to be this way.

But wouldn't it be possible to construct the url sever-sided and send it back in the json payload? So that the client can load and show the cached image.

kirillsalykin:
+1

tonklon:
Should be fixed by #440

PR #440, opened 2016-02-21
return an url to the just uploaded file in json

tonklon:
so that the client can access/display/download the file

see #429

### DIFF
--- a/lib/refile/app.rb
+++ b/lib/refile/app.rb
@@ -69,9 +69,11 @@ module Refile
     post "/:backend" do
       halt 404 unless upload_allowed?
       tempfile = request.params.fetch("file").fetch(:tempfile)
+      filename = request.params.fetch("file").fetch(:filename)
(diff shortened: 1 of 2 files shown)


choonkeat:
:+1: 

can consider returning the `content_type` and geometry (e.g. `200x400`) along with the "upload success" response e.g. https://github.com/choonkeat/attache#upload very useful for frontend

jnicklas:
@choonkeat interesting, hadn't seen attache. Refile can run as a separate server as well, by the way, it was one of the primary drivers behind the design. It isn't documented very well how to do it though ;)

Let's merge this, since it's definitely an improvement on the status quo. It's a bit unfortunate that we can't get a processed file out of this though, but there isn't really any generic way of doing this, I think.

tonklon:
@jnicklas I do have an idea how to return a process image. I will evaluate it and send a new pull request to discuss it.

PR #444, opened 2016-02-25
Readme for manual mounting apps with catch-all routes

tarr11:
Added some documentation if you use a catch-all route (for a CMS like [Comfortable Mexican Sofa](https://github.com/comfy/comfortable-mexican-sofa) ) which breaks Refile attachment routing


cerdiogenes:
+1

sobrinho:
Thanks ❤️

PR #448, opened 2016-03-01
forgot word

acandael:
I suppose it should be : "ActiveRecord integration ships with a nice macro that makes this trivial."


jnicklas:
Thanks!

PR #452, opened 2016-03-05
Resizing images before upload

cristhiano:
Gives support for max_width and max_height options for attachment_field helper so users can resize images before direct uploads

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,12 @@ cache: bundler
 
 sudo: false
 
+before_install:
+  - mkdir travis-phantomjs
+  - wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2
+  - tar -xvf $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 -C $PWD/travis-phantomjs
+  - export PATH=$PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH
+
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
@@ -25,4 +31,3 @@ matrix:
     - rvm: ruby-head
     - rvm: 2.2.4
       gemfile: gemfiles/Gemfile.rails-5.0-master
-
--- a/app/assets/javascripts/refile.js
+++ b/app/assets/javascripts/refile.js
@@ -19,7 +19,157 @@
     return data;
   }
 
+  function resizeParams(file, input) {
+    var info, srcRatio, trgRatio;
+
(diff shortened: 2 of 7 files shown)


90caiosantos:
Very Helpful, thanks!

chrishough:
+1 has this been merged?

90caiosantos:
not yet =/

jnicklas:
This is an impressive PR, but it's also a bit scary. There's a *lot* of code I don't quite understand yet, and it seems very complicated for what it's trying to do. I really like the general idea though, client side rescaling before upload would be *very* cool.

I'm a bit wary. I think the code needs to be cleaned up. There need to be tests. We need to ensure this works on every browser, on every platform. There is *a lot* of work in getting this PR merged.

cristhiano:
Hi @jnicklas, your right, it does need tests! I couldn't do it before but working on it now.

chrishough:
If you need help with cross browser testing I am game @cristhiano.  I could boot up VMs for fun.  I am pretty sure from past JS projects using this type of functionality it only works in modern browsers IE > = 9

cristhiano:
That will be great @chrishough! Your right, only modern browsers will do the job. I think that the big bottleneck is on Blob: http://caniuse.com/#search=Blob. IE 9 is out of picture because of this, but I don't think it's a problem since Refile it's already declared not compatible with IE 9. But maybe there is more problems, if you can clear this path will be awesome.

chrishough:
How about we add a comment in the docs that states this feature is only supported in Modern? Also, I would have to dig this up but there is a js lib that can be used to test for browser version and shut this off for safety.

cristhiano:
What about this: https://www.browserstack.com. It's free for open source projects.

chrishough:
I have used that before, it is not bad.  I would think if we test in Chrome, Firefox, Safari, and IE >= 10 we should be ok.  MSFT is pulling all support for the older browsers anyway.  Just my 2 pennies. Thoughts?

cristhiano:
I think that a Browserstack type of solution will cover us better when it comes to mobile devices. But in any case I will check for Blob constructing availability before rescaling images.

chrishough:
Sounds good.  @jnicklas  do you think we need anything else on this?

chrishough:
are we missing anything on this @jnicklas @cristhiano ?

olliekav:
@jnicklas @cristhiano Is this close to being merged before I write my own hacky implementation? :)

samuel012:
@cristhiano @chrishough any heads up as to when this will be merged?

chrishough:
@samuel012 I would defer to @jnicklas @cristhiano

sobrinho:
Hey, unfortunately, I'm going to close this one.

I don't think this belongs to Refile's core as you can use 3rd party javascript libraries for that and submit the resized file to the back-end deal with it later (just store in this case).

Since there are many options for that, I don't think one solution will be the answer for every user, I prefer to see Refile's core clean and small and delegate this to front-end teams.

PR #456, opened 2016-03-14
Document the 'autosave' option for multiple attachments

dwilkie:
Leaving out `autosave` will only save the attachments when the parent record is created

http://api.rubyonrails.org/classes/ActiveRecord/AutosaveAssociation.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,7 +664,7 @@ From the post model, you can use the `accepts_attachments_for` macro:
 
 ``` ruby
 class Post < ActiveRecord::Base
-  has_many :images, dependent: :destroy
+  has_many :images, dependent: :destroy, autosave: true
   accepts_attachments_for :images, attachment: :file
 end
 ```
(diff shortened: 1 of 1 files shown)


jnicklas:
Isn't that exactly what you want most of the time though?

dwilkie:
I don't think so. In the context of the README example, if your on the edit page for `Post` and you upload a new image and autosave is not set to true, then the image won't be saved. If autosave is set to true then by default the original image(s) will be replaced by the new uploads, or if `append: true` then the new images will be appended. I think this is the expected behavior.

dwilkie:
Another way to explain this is that when you use `accepts_nested_attributes_for` rails automatically sets the autosave option for you. http://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html

Since `accepts_attachments_for` works like `accepts_nested_attributes_for` to set the `files` attributes on the `images` model then autosave should be set to true for this association. Maybe we should do this automatically like in Rails?

PR #465, opened 2016-04-12
Passes the url parameter (added in #440) to the javascript

kindell:
Here is a suggestion to more convenient way of accessing the signed url when uploading a file.


sobrinho:
Thanks ❤️

PR #468, opened 2016-04-19
Add notice to Refile.secret_key

gabskoro:
We had a situation where we needed to change our `secret_key_base` due to a security policy and because we had some hard coded images (refile attachments) on our blog post everything broke.

After digging a while we realized that refile is using rails `secret_key_base` to generate to token, so it would be nice to have that written in the readme


sobrinho:
Thanks!

PR #479, opened 2016-05-25
Return presigned URL from S3

FabioMR:
Sometimes it is important to have de presigned URL from Amazon S3. 
It is much faster read direct from there.

### DIFF
--- a/lib/refile/attachment.rb
+++ b/lib/refile/attachment.rb
@@ -18,6 +18,7 @@ module Refile
     # - `remote_image_url`
     # - `remote_image_url=`
     # - `image_url`
+    # - `image_presigned_url`
     #
     # @example
     #   class User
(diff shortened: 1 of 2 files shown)


sobrinho:
Thanks ❤️

abitdodgy:
@FabioMR hi. Can you briefly explain what this pull request does and why it's useful? I'm just curious and I can't tell by the comment you left.

sobrinho:
I believe you can get the idea from here: https://blog.fyle.in/sharing-files-using-s3-pre-signed-urls-e05d4603e067

FabioMR:
@abitdodgy The article above explain better. Thanks @sobrinho!

sobrinho:
@FabioMR would be nice to receive another pull request including some documentation about how to use that feature with refile.

Can you do that?

PR #480, opened 2016-05-26
Make attachment_cache_field public

### LINKED ISSUE #391: Cache files between form submits without the input
When you submit a file and a validation error is presented, if you don't call `attachment_field` the file is lost.

I believe it is the hidden field generated by this method which keeps the cache between submits but I don't want to show the input again to the user if the file is already there, see the image.

screen shot 2015-11-03 at 14 42 35

What I've done is this:

``` erb
<tr class="nested-fields">
  <td>
    <%= f.input :id, as: :hidden %>

    <% if f.object.file %>
      <div style="display: none">
        <%= f.attachment_field :file %>
      </div>

      <%= link_to f.object.file_filename, attachment_url(f.object, :file) %>
    <% else %>
      <%= f.attachment_field :file %>
    <% end %>
  </td>

  <td class="text-right">
    <%= link_to_remove_association "Remover", f, class: 'btn btn-sm btn-danger' %>
  </td>
</tr>
```

Would be nice to have some kind of helper like `f.attachment_cache :file` to be able to keep the cache without rendering the file input itself.

Makes sense?

sobrinho:
Closes #391

### DIFF
--- a/lib/refile/rails/attachment_helper.rb
+++ b/lib/refile/rails/attachment_helper.rb
@@ -8,6 +8,12 @@ module Refile
         self.multipart = true
         @template.attachment_field(@object_name, method, objectify_options(options))
       end
+
+      # @see AttachmentHelper#attachment_cache_field
+      def attachment_cache_field(method, options = {})
+        self.multipart = true
+        @template.attachment_cache_field(@object_name, method, objectify_options(options))
+      end
     end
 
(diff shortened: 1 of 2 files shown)


sobrinho:
@jnicklas please, code review :)

jnicklas:
Thanks :) Looks good!

PR #483, opened 2016-06-10
Prevent rails (5) to add an extra hidden input

cristianbica:
Rails 5 adds a hidden input field for file fields: [commit here](https://github.com/rails/rails/commit/00b26532f05283d2b160308522d1bd2146d6ac18). This breaks refile's keeping the uploaded file across form redisplay (like other fields validations). This PR fixes that by passing the `include_hidden` to the file_filed call.

### DIFF
--- a/lib/refile/rails/attachment_helper.rb
+++ b/lib/refile/rails/attachment_helper.rb
@@ -83,6 +83,7 @@ module Refile
       end
 
       options[:data][:reference] = SecureRandom.hex
+      options[:include_hidden] = false
 
       hidden_options = {
         multiple: options[:multiple],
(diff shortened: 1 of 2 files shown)


cristianbica:
It seems that this fixed all the rails 5 failure (there were 4 or 5)

cristianbica:
For who's interested my current workaround (I'm using simple_form)
```ruby
f.input :photo, as: :attachment, input_html: { include_hidden: false }
```

jnicklas:
Ugh, that addition in Rails is just wrong. This is not how file fields work :(

cristianbica:
Yes. While I understand a bit the reasoning the behavior should be reversed as it very rare when you have a form with just one photo

-- 
Cristian Bica

PR #486, opened 2016-07-01
Update gemspec to support mime-types v3

jastkand:


artofhuman:
What the status for this pr?
@jnicklas /cc

manfe:
Sinatra in version `1.4.5` and `1.4.6` has rack dependency in version `~> 1.4` and for version `1.4.7` is the rack version dependency is `~> 1.5` which means that in rails 5 this is not going to work.

Because of this: 
```
actionpack (= 5.0.0.1) was resolved to 5.0.0.1, which depends on
          rack (~> 2.0)
```
So the correct thing to do is this. https://github.com/manfe/refile/commit/5ad97dda441247287631969a12ed6be0b9417d94

jastkand:
@manfe I'm using this implementation on Rails 5, so it does work for me. In order to make it work I had to specify sinatra version explicitly in my `Gemfile`.

```ruby
gem 'sinatra', '~> 2.0.0.beta2'
gem 'rack-protection', '~> 2.0.0.beta2' # rack-protection is also required for sinatra 2.0
```

I think it's a better approach, because I don't like the idea of specifying beta version of dependency in the gem itself.

manfe:
@jastkand You are right. Adding `2.0.0.beta2` directly as dependency broke refile for older versions of ruby. So it's better to do what you suggested.

onebree:
@jastkand I am using Rails 5 as well. I had some dependency issues when installing refile from RubyGems, so I instead specified Github. The following is shown when I run `bundle install`

```
Bundler could not find compatible versions for gem "mime-types":
  In snapshot (Gemfile.lock):
    mime-types (= 3.1)

  In Gemfile:
    rails (>= 5.0.0.1, ~> 5.0.0) was resolved to 5.0.0.1, which depends on
      actionmailer (= 5.0.0.1) was resolved to 5.0.0.1, which depends on
        mail (>= 2.5.4, ~> 2.5) was resolved to 2.6.4, which depends on
          mime-types (< 4, >= 1.16)

    refile was resolved to 0.6.2, which depends on
      mime-types

    refile was resolved to 0.6.2, which depends on
      rest-client (~> 1.8) was resolved to 1.8.0, which depends on
        mime-types (< 3.0, >= 1.16)
``` 

`rest-client` version 2.0.0 accepts `mime-types` up to version `< 4.0`

jastkand:
@onebree, my `Gemfile` looks this way:

```
gem 'rest-client', '>= 2.0.0.rc4'
gem 'sinatra', '~> 2.0.0.beta2'
gem 'refile', require: 'refile/rails', git: 'https://github.com/jastkand/refile.git', branch: 'mime-types-v3'
```

thomasfedb:
This is a good interim solution. Could it be merged?

havenwood:
It'd be lovely to allow rest-client 2.0.

For what it's worth, Sinatra 2.0.0.rc2 has been released.

sobrinho:
We are working on this at #529. Thanks!

PR #494, opened 2016-07-28
Add possibility to specify attachment extensions via Proc

### LINKED ISSUE #493: Add possibility to specify attachment extensions via Proc
Currently attachment extension constraints are set up in application initialization phase but it would be nice to be able to change them while application is still running.
For example web page administrator could change list of whitelisted extensions in admin panel or database.

This could be solved by allowing to specify extensions via Ruby Proc

``` ruby
attachment :cv, extension: -> { Settings.find_by(code: "cv_extension_whitelist").value }
```

I will create a pull-request with working implementation

krists:
Fixes #493

### DIFF
--- a/lib/refile/attachment_definition.rb
+++ b/lib/refile/attachment_definition.rb
@@ -1,7 +1,7 @@
 module Refile
   # @api private
   class AttachmentDefinition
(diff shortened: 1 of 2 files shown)


sobrinho:
Thanks, this open a few interesting features for the refile.

PR #503, opened 2016-08-17
Fix multibyte filename when file downloading on Firefox 48.0.

bundai223:
When download uploaded file that named by multibyte charactors（like "てすと.pdf" in Japanese） on Firefox,
downloaded file name is garbled（like "%E3%81%A6%E3%81%99%E3%81%A8.pdf"）.

### DIFF
--- a/lib/refile/app.rb
+++ b/lib/refile/app.rb
@@ -1,6 +1,7 @@
 require "json"
 require "sinatra/base"
 require "tempfile"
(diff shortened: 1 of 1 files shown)


sobrinho:
Can you make some specs? <3

sobrinho:
Fixed on a53308a12f0ba8bc8e473e3513749afc311f9ff3.

Issue #447, opened 2016-03-01
Can't install on Rails 5


arogachev:
I want to use Refile for AJAX file upload in app based on Rails 5.0.0.beta3. I added this to Gemfile as mentioned in docs:

``` ruby
gem 'refile', require: 'refile/rails'
gem 'refile-mini_magick'
```

Then after running `bundle install` I get the following error:

```
Fetching gem metadata from https://rubygems.org/.............
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies.......
Bundler could not find compatible versions for gem "rack":
  In snapshot (Gemfile.lock):
    rack (= 2.0.0.alpha)

  In Gemfile:
    unicorn (>= 0) ruby depends on
      rack (>= 0) ruby

    rails (< 5.1, >= 5.0.0.beta1) ruby depends on
      railties (= 5.0.0.beta3) ruby depends on
        actionpack (= 5.0.0.beta3) ruby depends on
          rack (~> 2.x) ruby

    sass-rails (>= 3.2) ruby depends on
      sprockets (< 4.0, >= 2.8) ruby depends on
        rack (< 3, > 1) ruby

    rails (< 5.1, >= 5.0.0.beta1) ruby depends on
      railties (= 5.0.0.beta3) ruby depends on
        actionpack (= 5.0.0.beta3) ruby depends on
          rack-test (~> 0.6.3) ruby depends on
            rack (>= 1.0) ruby

    rails (< 5.1, >= 5.0.0.beta1) ruby depends on
      railties (= 5.0.0.beta3) ruby depends on
        actionpack (= 5.0.0.beta3) ruby depends on
          rack-test (~> 0.6.3) ruby depends on
            rack (>= 1.0) ruby

    refile-mini_magick (>= 0) ruby depends on
      refile (~> 0.5) ruby depends on
        sinatra (~> 1.4.5) ruby depends on
          rack (~> 1.4) ruby

    unicorn (>= 0) ruby depends on
      rack (>= 0) ruby

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

`bundle update` didn't help. I also tried to delete `Gemfile.lock` and run `bundle install` again, didn't help also.

It's important to mention that I included latest version of rack in my Gemfile, because of this error - https://github.com/activeadmin/activeadmin/issues/4327:

``` ruby
gem 'rack', github: 'rack/rack'
```

[...]

PR #508, opened 2016-08-31
Add support to rails 5

manfe:
This is related to issue #447 and PR #486.

In #447 I made a full log of what I did, also commented in #486, since it will fix mime-types but will not work in rails 5.

Also this PR, is not to merge yet. We should wait [sinatra](https://github.com/sinatra/sinatra) release the 2.0 stable version.

Thanks

### DIFF
--- a/refile.gemspec
+++ b/refile.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.1.0"
 
-  spec.add_dependency "rest-client", "~> 1.8"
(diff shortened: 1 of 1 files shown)


manfe:
I am closing this PR, since this is not compatible with previous versions of rails.

I tried to test it locally, but no success.

So the best solution is to use `gem "sinatra", "~> 2.0"` instead of putting the version as dependent, otherwise will break older version of rails.

hwyte:
I am getting this error trying to include sinatra as well. 

Could not find gem 'sinatra (~> 2.0)' in any of the gem sources listed in your Gemfile or available on this machine.

PR #518, opened 2016-12-03
Feature: expirable download links

gkop:
This is a quick and dirty implementation of expirable download links.

The use case is that we have sensitive attachments for which long-lived download links are unacceptable.

I'd love to get some feedback :)


sobrinho:
But if the user has the URL and simple remove the `expires_at`, the link will continue to work, no?

gkop:
Gabriel, I see how you could get that understanding from the diff, but in fact this case is accounted for as `expires_at` is included in what is signed for the signature. I will triple check this and write a test case...

sobrinho:
Merged, thanks!

PR #526, opened 2017-03-01
Prevent from exposing secure information unexpectedly

AknEp:
I'm using refile with Amazon S3.

I found `Refile::File` class potentially exposes important backend information (includes AWS Secret Key! )

```ruby
# entity.rb
class Entity < ApplicationRecord
  attachment :image
end

# in jbuilder view, @entity is Entity's instance
json.image @entity.image
```

or, `@entity.image.as_json` in rails console.

Although I know this jbuilder code is wrong, but someone might writes this code.

I think important information should not revealed with trivial human error.

### DIFF
--- a/lib/refile/file.rb
+++ b/lib/refile/file.rb
@@ -83,6 +83,16 @@ module Refile
       @io = nil
     end
 
+    # Prevent from exposing secure information unexpectedly
+    #
+    # @return [Hash]
+    def as_json(options={})
+      {
+        id: id,
+        backend: backend.to_s
+      }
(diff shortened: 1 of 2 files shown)


sobrinho:
Hi! Revert the gemspec change and rebase against master and add some spec.

I can provide you tips on how to do that if you need, feel free to ping me.

lucasdeassis:
Continued this PR in #562 adding some specs.

sobrinho:
Closing in favor of #562 .

PR #546, opened 2017-10-31
Disallow zero byte file upload

peiyee:
I am using Refile and [Minimagick](https://github.com/minimagick/minimagick) to resize my images on the fly, and sometime our system admin will accidentally upload zero byte file to our server. Then minimagick will raise error when it wants to resize the zero byte file.

Because of this, this PR adds a new option to validate the file size is zero byte or not. Set `allow_zero_byte` to `false` will raise error when users try upload zero byte file to the server.

@sobrinho please let me know this new feature is good to go, so that i will update the README. Thanks!!!

### DIFF
--- a/lib/refile/attachment_definition.rb
+++ b/lib/refile/attachment_definition.rb
@@ -50,6 +50,7 @@ module Refile
       errors << :invalid_extension if valid_extensions and not extension_included
       errors << :invalid_content_type if valid_content_types and not valid_content_types.include?(attacher.content_type)
       errors << :too_large if cache.max_size and attacher.size and attacher.size >= cache.max_size
+      errors << :zero_byte_detected if attacher.size.to_i.zero?
       errors
     end
   end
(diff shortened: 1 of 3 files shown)


sobrinho:
@peiyee hey, good to see you sending PRs here!

Any idea of a use-case for zero byte files?

I'm thinking in not allowing it at all, it should be always a mistake 🤔

peiyee:
@sobrinho, i cannot think of any use-case as well. I agree to not allowing it at all. 
Shall i add this validation by editing this PR or create a new one?

sobrinho:
@peiyee you can edit this one 👍

peiyee:
@sobrinho i have updated my PR to not allow zero byte file upload, please help to review. 😄

sobrinho:
@peiyee merged, seems perfect, thanks!

manuelpuyol:
Well, there might be an use-case..
If you are using `Tempfiles` they start as a 0 byte and after it is uploaded, they act as a normal file on the filesystem.

Before this change, I could upload a `Tempfile` normally (also download it without any problems), now it blocks be

Issue #370, opened 2015-09-24
Assume the existence of `"#{name}_id_will_change!"` in a class that inherits from ActiveRecord::Base can cause problems


cerdiogenes:
I'm using ActiveType and some objects defined by it inherits from ActiveRecord::Base, but doesn't implement the ActiveModel::Dirty API, because they aren't persisted to the database.

While trying to use refile with it I get the error `undefined method 'blah_id_will_change!'`. Initially I opened  an issue against ActiveType: https://github.com/makandra/active_type/issues/50

For they it doesn't make sense to implement ActiveModel::Dirty API, since the data isn't persisted in a database, so it's a won't fix.

They suggested a check with `respond_to?("#{name}_id_will_change!")` [here](https://github.com/refile/refile/blob/20ecdb3d760b8eed2d5f4bddf6dd8c673063bf2c/lib/refile/attachment/active_record.rb#L26): https://github.com/makandra/active_type/issues/50#issuecomment-142463371

This is a reasonable solution @jnicklas?

jnicklas:
Yes, that sounds reasonable to me.

sobrinho:
Closed by #558.

PR #558, opened 2018-01-16
Allow assign file to non trackable attribute

marcelobalexandre:
I do not know if this correction is still wanted.

Fixes an error that occurs when trying to assign a file to an attribute that does not track changes or which class inherits from ActiveRecord::Base but does not implement ActiveModel::Dirty.

https://github.com/refile/refile/issues/370


sobrinho:
Thanks ❤️

PR #562, opened 2018-02-19
Prevent from exposing secure information unexpectedly - specs

lucasdeassis:
Continued PR #526. Added specs.

### DIFF
--- a/spec/refile/backend_examples.rb
+++ b/spec/refile/backend_examples.rb
@@ -221,8 +221,7 @@ RSpec.shared_examples_for :backend do
         file = backend.upload(uploadable)
         hash = file.as_json
 
-        expect(hash).to include(:backend)
-        expect(hash[:backend]).to be_a(String)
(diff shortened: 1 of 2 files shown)


spec/refile/backend_examples.rb
```diff
@@ -215,5 +215,15 @@ def uploadable(data = "hello")
...
       end
     end
+
+    describe "#as_json" do
+      it "returns id and stringified backend as a Hash" do
+        file = backend.upload(uploadable)
+        hash = file.as_json
+
+        expect(hash).to include(:backend)
+        expect(hash[:backend]).to be_a(String)
```
sobrinho: Assert that only the keys id and backend are exposed, this will guarantee that no-one adds extra information unexpected to this method.

Something like:

```ruby
expect(hash.keys).to eq [:id, :backend]
```

Instead of the two assertions.

lucasdeassis: change applied.

PR #564, opened 2018-02-27
Custom msg format validation

elmasantos:
This PR implements the custom validation message, which return the invalid extension or type that the user is trying to upload and shows the allowed ones.

### DIFF
--- a/lib/refile/attachment_definition.rb
+++ b/lib/refile/attachment_definition.rb
@@ -45,13 +45,38 @@ module Refile
     end
 
     def validate(attacher)
+      extension = attacher.extension.to_s.downcase
+      content_type = attacher.content_type.to_s.downcase
+
       errors = []
-      extension_included = valid_extensions && valid_extensions.map(&:downcase).include?(attacher.extension.to_s.downcase)
-      errors << :invalid_extension if valid_extensions and not extension_included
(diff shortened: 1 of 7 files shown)


lib/refile/attachment_definition.rb
```diff
@@ -45,13 +45,38 @@ def valid_extensions
...
+      errors << content_type_error_params(content_type) if invalid_content_type?(content_type)
       errors << :too_large if cache.max_size and attacher.size and attacher.size >= cache.max_size
       errors << :zero_byte_detected if attacher.size.to_i.zero?
       errors
     end
+
+  private
+
+    def extension_error_params(extension)
+      [:invalid_extension, extension: format_param(extension), permitted: valid_extensions.join(", ")]
```
sobrinho: Prefer to use `valid_extensions.to_sentence` because the join argument may differ for different languages (I18n).

lib/refile/attachment_definition.rb
```diff
@@ -45,13 +45,38 @@ def valid_extensions
...
+    end
+
+    def content_type_error_params(content_type)
+      [:invalid_content_type, content: format_param(content_type), permitted: valid_content_types.join(", ")]
```
sobrinho: Same here.

Issue #431, opened 2016-02-03
Image validation fails on valid images


andrewhavens:
Not sure which of these is the problem. But I'm finding that some valid images are not passing validation. I think it might be a mime type mismatch on the server, or maybe it's the inclusion of `charset=UTF8`. For example:

In my model:

``` ruby
attachment :attachment, type: :image
```
## Example 1:
- Filename: `.jpg`
- Mime type: `image/jpeg`
- Response header: `Content-Type: image/png;charset=UTF-8`
- URL: http://blogs.scientificamerican.com/blogs/cache/file/8197CEC8-990D-4905-AE984685E38CA6BF_agenda.jpg?w=600&h=335
## Example 2:
- Filename: `.png`
- Mime type: `image/jpeg`
- Response header: `Content-Type: image/jpeg; charset=utf-8`
- URL: https://www.asus.com/media/global/products/tlvlZTN8Y0VoKyiB/P_setting_fff_1_90_end_600.png

Maxoos:
I've seen the same, not sure why. I don't think this rather excellent gem is maintained anymore or as often as needed.

sobrinho:
I'm not seeing this issue by using `remote_{attachment}_url` or downloading it and uploading through the browser.

**Example 1:**

``` bash
λ curl -L 'http://blogs.scientificamerican.com/blogs/cache/file/8197CEC8-990D-4905-AE984685E38CA6BF_agenda.jpg?w=600&h=335' | md5
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 25249  100 25249    0     0   163k      0 --:--:-- --:--:-- --:--:--  163k
a5a3fa6e967a6ee4606d00b7ca95fba8
```

```ruby
[15] pry(main)> x = User.create(remote_avatar_url: 'http://blogs.scientificamerican.com/blogs/cache/file/8197CEC8-990D-4905-AE984685E38CA6BF_agenda.jpg?w=600&h=335')
   (0.1ms)  begin transaction
[...]

PR #569, opened 2018-03-13
Fix 'validate' method to ignore the encoding part in content type

elmasantos:
This PR fixes the `Attacher#content_type` method, in a way that if an **encoding** is appended to content-type response header, now it will be ignored and for now on valid images will pass the validation.

#431

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -38,8 +38,7 @@ module Refile
     end
 
     def content_type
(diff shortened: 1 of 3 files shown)


sobrinho:
What's the impact on other content types that are allowed to have an encoding defined? Does Refile today save the encoding or the encoding is never sent?

elmasantos:
Well, it doesn't seem to have impact since Refile only looks for the type and subtype specified in the content-type. Parameters like "charset" are not considered.

Issue #304, opened 2015-07-22
undefined method `files_attachment_definition'


DerAlSem:
Hi,

Trying to switch from Paperclip to Refile.

`gemfile` (just in case)

``` ruby
gem "refile", require: "refile/rails", git: 'https://github.com/refile/refile.git', branch: 'master'
gem "refile-mini_magick"
```

I have features:

`feature.rb`

``` ruby
class Feature < ActiveRecord::Base
    has_many :assets, dependent: :destroy
    accepts_attachments_for :assets, append: true  

    def post_params
      params.require(:feature).permit(files: [])
    end
end
```

Each features has assets. 

`asset.rb`

``` ruby
class Asset < ActiveRecord::Base
    belongs_to :feature
    attachment :file

    enum category: { 'А/В' => 1, 'Для прессы' => 2, 'Документы' => 3,  'Другое' => 4 }
    categories
    [['A/V', 1], ['Для прессы', 2], ['Документы', 3]]

    def asset_params
        params.require(:asset).permit(:file)
    end

end
```

Basically an asset - is a file, which can be downloaded. Before downloading it should be uploaded. Hi, captain. So, form is:

`views/assets/form.html.erb`

``` ruby
<%= form_for @asset, :url => {:action => :create, :method => :post}, :html => {:multipart => true} do |f| %>
 <p>
    <strong>Choose file for upload</strong>
    <%= f.attachment_field :files, multiple: true %>
  </p>
end
```

NoMethodError in Assets#new 

undefined method `files_attachment_definition' for #Asset:0xb41c35a4

for line
`<%= f.attachment_field :files, multiple: true %>`

I've run through docs once more. Noticed, that controller should be updated too.

``` ruby
def feature_params
params.require(:feature).permit(assets_files: [], :title_en)
end 
```

This gives me another error: 

SyntaxError in FeaturesController#edit 
/home/deralsem/devel/upi-rus/app/controllers/features_controller.rb:72: syntax error, unexpected ',', expecting => ...it(assets_files: [], :title_en, :title_ru, :release_date, :g... ... ^

Any advise much appreciated.

Update:

``` ruby
def feature_params
params.require(:feature).permit(:assets_files [], :title_en)
end 
```

moved colon before asset_files, error is gone, but undefined_method is still there. ;-(

abitdodgy:
You assert model:

```ruby
categories
[['A/V', 1], ['Для прессы', 2], ['Документы', 3]]
```

[...]

PR #570, opened 2018-03-13
Add message to wrong association name

JuarezLustosa:
Fix issue #304


config/locales/en.yml
```diff
@@ -7,4 +7,5 @@ en:
         invalid_content_type: "You are not allowed to upload %{content} file format. Allowed types: %{permitted}."
         invalid_extension: "You are not allowed to upload %{extension} file extension. Allowed types: %{permitted}."
   refile:
-    empty_param: "an empty"
\ No newline at end of file
+    empty_param: "an empty"
+    wrong_association_method: "wrong association name %{method}, use like this %{sugest}"
```
sobrinho: Not necessary to use i18n here, it is a programmer error, just use a string directly on lib/refile/attachment/active_record.rb.

spec/refile/attachment/active_record_spec.rb
```diff
@@ -331,6 +331,19 @@ def self.name
     let(:post) { post_class.new }
 
     describe "#:association_:name" do
+      let(:wrong_method) { "files" }
+      let(:wrong_association_message) do
+        "wrong association name #{wrong_method}, use like this documents_files"
+      end
+
+      it "return erro for wrong association name" do
```
sobrinho: Use `returns a friendly error message for wrong association name`.

grosser:
so how about a release ... last is from 2015 :trollface:

Issue #446, opened 2016-02-29
Multiple files upload with pure ruby classes 


andersonfernandes:
Hi guys,

I need to upload multiple files with pure ruby classes, but i'm having some problems because the [solution](https://github.com/refile/refile#multiple-file-uploads) pointed on the documentation use `accepts_attachments_for`, and the object needs to be an activerecord.

Someone can help me with that?

arogachev:
@andersonfernandes +1 for that.

In my case I have content form in modal and I want to add file attachments to it. But since form is inside of a modal, file uploading should be AJAX / jQuery based.

I don't have ActiveRecord model and don't want to save anything to database. I use https://github.com/plataformatec/mail_form and https://github.com/plataformatec/simple_form. Provided examples contains usage of attachment with ActiveRecord model. Any tips on how to setup that?

sobrinho:
I think that won't be easy to do with the accepts_attachments_for macro right now because it relies on the AR associations for that.

But a similar macro should not be too hard to do, looking for suggestions on this.

sobrinho:
@elmasantos will chime on this one.

thiagoa:
Basically, we would need the client to implement an interface. For example:

```ruby
class Image
  extend Refile::Attachment

  attachment :file
end

class User
  extend Refile::Attachment

  attr_reader :images

  accepts_attachments_for :images, collection_class: Image
end
```

thiagoa:
@sobrinho @elmasantos It would be great if we could flesh this out before implementing.

We have two dependencies on ActiveRecord:

- `reflect_on_association`
- `ActiveRecord::Relation#build`

`reflect_on_association` is used to figure out the ActiveRecord association class. With an extra `collection_class` parameter, we could easily solve the problem for the non-AR case. For example:

```ruby
accepts_attachments_for :images, collection_class: Image
```

Now onto `ActiveRecord::Relation#build`. The AR code is similar to:

```ruby
send(association_name).build(attachment => file)
```

The same functionality could be ported to a non-ActiveRecord context by doing something like:

```ruby
send(collection_name) << collection_class.new(attachment => file)
```

To make it work, we would need to trust the client to implement an interface:

- An `attr_reader :images`
- An ActiveRecord-like constructor

Take the following example:

```ruby
class Image
  extend Refile::Attachment

  attachment :file
end

class User
  extend Refile::Attachment

  attr_reader :images

  accepts_attachments_for :images, collection_class: Image

  def initialize(attributes = {})
[...]

PR #571, opened 2018-03-27
Support multiple file upload in POROs

elmasantos:
This pull request implements a new `accepts_attachments_for` macro that works with pure Ruby classes. Previously, this macro only worked with Active Record classes. I also kept the Active Record macro.

To accomplish that:

- Macro code has been extracted to a new method called `Refile::Attachment::MultipleUploads#new`. This method generates a module that can be used by both versions of `accepts_attachments_for`.
- To use a PORO, users of the gem will need to implement an interface that's similar to Active Record's. See the README.md file for more details.
- I've moved macro specs to an RSpec shared example and also improved the test coverage.

Also, I had to change the valid `expires_at` from 1 minute to 1 second in the `app_spec.rb` file, in order to avoid CI build failing.

#446

### DIFF
--- a/README.md
+++ b/README.md
@@ -784,6 +784,69 @@ class Post < ActiveRecord::Base
   accepts_attachments_for :images, append: true
 end
 ```
+### Multiple file uploads for pure Ruby classes
+
+You can also use `accepts_attachments_for` macro in pure Ruby classes.
+
+Suppose you have a `Document` class to be associated with a `Post` class.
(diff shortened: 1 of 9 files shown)


elmasantos:
@sobrinho my branch is now up to date with master and has no conflicts. Could you check it again and see if it's ok to merge?

sobrinho:
Merged, thanks!

sobrinho:
s2

Issue #555, opened 2017-12-31
Refile JS as an NPM package


bobbytables:
In my project I'm not using the Rails asset pipeline. I'm trying to use `presigned` for my s3 uploads but realized it requires the presence of the Javascript library. It'd be nice to import this as an NPM package in my entry file.

sobrinho:
I would like to see a PR implementing that, just do not publish the package (I or someone else from the Refile's team needs to be in control of that) but I would like to see the automation part done in a PR.

Until then, you can copy the src from the GitHub.

PR #575, opened 2018-07-10
Support Publishing JS as an NPM Package

tristanoneil:
This would close out #555 (if it gets pulled in). I published this @ https://www.npmjs.com/package/testing-refile-tristanoneil for testing purposes so you can try it out with `npm install --save testing-refile-tristanoneil`. The package only includes `refile.js` in addition to `README.md`, `History.md`, `LICENSE.txt` and `package.json` (`npm` publishes these files by default). It does appear that [`refile`](https://www.npmjs.com/package/refile) is already taken on NPM so the package name would likely have to be changed, I'll leave that up to the maintainer though as to what it should be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,9 @@ simply include it like this:
 //= require refile
 ```
 
+You can also separately install the JavaScript library with NPM or Yarn using
+`npm install --save refile` or `yarn add refile`.
+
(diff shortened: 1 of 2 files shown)


tristanoneil:
Travis CI failure seems to be unrelated.

tristanoneil:
@sobrinho any thoughts?

PR #578, opened 2018-10-22
Allow assignment of attachment that was previously nil

Jcambass:
Closes: https://github.com/refile/refile/issues/579

Currently it's not possible to assign a attachment that was previously
set to `nil` without reinitializing the object.

Previously the following code would not work:

```ruby
post.document = nil
post.save
post.document = Refile::DoubleFile.new("hello")
post.save

post.document # => still nil
```

This bug was caused by the `remove` attribute which was set to `true`
when `nil` was assigned but never reset to `false` when a valid
attachment was assigned afterwards.

The previous workaround was to re-create the object eq. calling
something like `post = Post.find(post.id)`

### DIFF
--- a/lib/refile/attacher.rb
+++ b/lib/refile/attacher.rb
@@ -73,6 +73,7 @@ module Refile
     end
 
     def set(value)
+      self.remove = false
       case value
         when nil then self.remove = true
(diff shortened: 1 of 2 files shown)


sobrinho:
Thanks! <3

Issue #348, opened 2015-09-08
Error when not using `multiple: true` in attachment_field


abitdodgy:
In my application I have an automatic versioning feature where a `Document` object has many `Version` objects. From the `documents#show` I want to enable the user to upload a new version of the document. 

``` ruby
class Document < ActiveRecord::Base
  accepts_attachments_for :versions, append: true
end
```

Here's the form.

``` slim
  = form_for [@project, @document] do |f|
    = f.attachment_field :versions_files, presigned: true, host: request.base_url
    = f.submit
```

In this case I want the user to upload one file at a time, so I decided to omit the `multiple: true` option. However, the object was not being updated. After looking at the logs, I found that `strong_parameters` was filtering the `versions_files` param.

```
Unpermitted parameter: versions_files
```

But the parameter is whitelisted, so I figured its type validation is failing.

``` ruby
# documents_controller.rb
@document.update! permitted_attributes(@document)
```

``` ruby
class DocumentPolicy < ApplicationPolicy
  def permitted_attributes
      [:status, :phase, :file, versions_files: []]
  end
end
```

I changed the `permitted_attributes` method to return `versions_files` as a symbol.

```
  def permitted_attributes
      [:status, :phase, :file, :versions_files]
```

Then I got this error:

```
ArgumentError in DocumentsController#update
```

From the stack trace, it looks like it's happening in `accepts_attachments_for`... I've omitted much of the stack trace because I don't think it's useful. I'll add it if it's needed.

```
/Users/Mohamad/.rvm/gems/ruby-2.2.0/bundler/gems/refile-973d4ad5617a/lib/refile/attachment/active_record.rb:99:in `partition'
/Users/Mohamad/.rvm/gems/ruby-2.2.0/bundler/gems/refile-973d4ad5617a/lib/refile/attachment/active_record.rb:99:in `block (2 levels) in accepts_attachments_for'
activerecord (4.2.3) lib/active_record/attribute_assignment.rb:54:in `public_send'
activerecord (4.2.3) lib/active_record/attribute_assignment.rb:54:in `_assign_attribute'
```

I'm using master.

jnicklas:
I think that Refile assumes that `versions_file` is an array, whereas from what I understand in your case it wouldn't be. IMO this case should be made to work. A PR would be fantastic!

sobrinho:
I'm not sure about this one...

`versions_files` expects an array and accepting a File object is strange to me, although I see the use case.

@abitdodgy want to open a PR for that one?

[...]

PR #584, opened 2019-01-09
Allow support for single file upload

tiarly:
It closes https://github.com/refile/refile/issues/348

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Features:
 - Works across form redisplays, i.e. when validations fail, even on S3
 - Effortless direct uploads, even to S3
 - Support for multiple file uploads
+- Support for single file upload
 
 Sponsored by:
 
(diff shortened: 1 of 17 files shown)


tiarly:
The failing CI build isn't really related to this pull request code 😉

README.md
```diff
@@ -784,6 +785,62 @@ class Post < ActiveRecord::Base
...
+attribute:
+
+``` erb
+<%= form_for @post do |form| %>
+  <%= form.label :images_files %>
+  <%= form.attachment_field :images_files %>
+<% end %>
+```
+
+Now you only need to permit the generated accessor in your controller.  Since
```
abnersajr:
```suggestion
Now you only need to permit the generated accessor in your controller. Since
```

sobrinho:
@tiarly can you open another PR fixing that build problem?

I can't merge a broken build, even when it is not the root cause.

tiarly:
> @tiarly can you open another PR fixing that build problem?
> 
> I can't merge a broken build, even when it is not the root cause.

Will do!

sobrinho:
Thanks! <3

PR #585, opened 2019-03-08
Changed deprecated method

VitorCarminatti:
I was recieving annoyng deprecation warnings while running my tests, so I decided to fork and update the method

### DIFF
--- a/lib/refile/attachment/active_record.rb
+++ b/lib/refile/attachment/active_record.rb
@@ -24,17 +24,17 @@ module Refile
         end
 
         define_method "#{name}=" do |value|
-          send("#{name}_id_will_change!") if respond_to?("#{name}_id_will_change!")
+          send("will_save_change_to_#{name}_id?") if respond_to?("#{name}_id_will_change!")
           super(value)
         end
 
         define_method "remove_#{name}=" do |value|
-          send("#{name}_id_will_change!")
+          send("will_save_change_to_#{name}_id?")
           super(value)
         end
 
         define_method "remote_#{name}_url=" do |value|
-          send("#{name}_id_will_change!")
+          send("will_save_change_to_#{name}_id?")
           super(value)
         end
 


sobrinho:
Hello! Can you rebase that against master?

sobrinho:
Apparently, we can't just change because we support some old versions as well.

You probably need to check if the object responds to the method before calling it.

Can you do that?

The CI is showing the error.

JuarezLustosa:
I think this pull request is unnecessary

I've tested this with 3 versions of rails,  5.2.3, 5.2.2 and 5.2.1, and they didn't produce any warning about it. 
I guess perhaps this happened due to of this [issue](https://github.com/rails/rails/issues/31087) on Rails.
This method "will_change!" in ActiveRecord::Dirty was showing a deprecation warning because of its call "attribute_changed?" inside of the method, and now it seems to be fixed 

Then I think we can close this pull request.

Captura de Tela 2019-04-10 às 18 04 43

sobrinho:
Thanks!

lib/refile/attachment/active_record.rb
```diff
@@ -24,17 +24,17 @@ def attachment(name, raise_errors: false, destroy: true, **options)
         end
 
         define_method "#{name}=" do |value|
-          send("#{name}_id_will_change!") if respond_to?("#{name}_id_will_change!")
+          send("will_save_change_to_#{name}_id?") if respond_to?("#{name}_id_will_change!")
```
mathieujobin, line 27: I don't believe this is equivalent.

the bang method changes the value
the ? only returns a boolean

```
(byebug) will_save_change_to_review_category_cache?
false
(byebug) review_category_cache_will_change!
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead. (called from refresh_state! at (byebug):1)
#<Set: {"review_category_cache"}>
(byebug) will_save_change_to_review_category_cache?
true
(byebug) 
```

PR #589, opened 2019-04-12
Bump version

KimEmmanuel:
Change version from 0.6.2 to 0.7.0 because we can't differ issues using master and the released version


sobrinho:
Thanks ❤️

PR #591, opened 2019-05-17
chore: simple fix in the documentation for active record's accepts_attachments_for

ristovskiv:
This is a very simple fix, but it can cause confusion while reading the source code or using docs.


sobrinho:
Thanks! <3

ristovskiv:
@sobrinho it was a pleasure

PR #596, opened 2019-10-03
Support Rails Credentials

Jcambass:
This patch adds support for the credentials feature introduced in rails 5.2.

The code patch has been updated to also check that the `secret_key_base` is also actually set. This is more in-line with the logic of `config.secret_key_base` in [Rails 5.2](https://api.rubyonrails.org/v5.2.1/classes/Rails/Application.html#method-i-secret_key_base) and whit what other gems like [Devise](https://github.com/plataformatec/devise/blob/master/lib/devise/secret_key_finder.rb) do.

We can't use `config.secret_key_base` since it doesn't exist on Rails versions before 5.2.

Replaces https://github.com/refile/refile/pull/595
Closes https://github.com/refile/refile/issues/593

### DIFF
--- a/lib/refile/rails.rb
+++ b/lib/refile/rails.rb
@@ -27,13 +27,21 @@ module Refile
     end
 
     initializer "refile.secret_key" do |app|
-      Refile.secret_key ||= if app.respond_to?(:secrets)
+      Refile.secret_key ||= if app.respond_to?(:credentials) && key_exists?(app.credentials)
+        app.credentials.secret_key_base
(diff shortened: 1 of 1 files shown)


sobrinho:
@Jcambass restarting the work in Refile.

Can you add some specs for this? Maybe add the newer rails version to CI to make sure that it works?

Jcambass:
For some reason I've only seen your message now :/ Unfortunately I don't use Refile anymore so I'm going to close this PR. Feel free to piggy back on this PR.

PR #603, opened 2021-01-14
Support Ruby 3

vitalinfo:


alexplatteeuw:
Hello, is the pull request going to be merged and the version bumped?

willkoehler:
Thanks @vitalinfo for your work on Ruby 3.x support 🙌 This has been a big help to unblock our Rails 7 + Ruby 3.2 upgrade while we look for options to migrate off refile.

vitalinfo:
@willkoehler np, if you have any roadmap/ideas to migrate off refile will be thankful for your thoughts

alexplatteeuw:
@vitalinfo I also thank you for your work, it also helped us while migrating off from refile (& refile-s3) to active storage 🙏

willkoehler:
> @willkoehler np, if you have any roadmap/ideas to migrate off refile will be thankful for your thoughts

Updating an application to use `ActiveStorage` (or in our case `CarrierWave` + S3) is pretty straightforward. The biggest challenge for us will be migrating existing attachments out of PostgreSQL to S3 (the app currently uses `Refile::Postgres` 🤦‍♂️). My plan is to use a Rake task, loop through each attachment and migrate them individually. I'll circle back here when I have something working.

berkos:
Thanks so much for providing this PR. It would be great if we can cut a new gem version with these changes. 🙏  🙏

lib/refile/attachment.rb
```diff
@@ -85,8 +85,8 @@ def attachment(name, cache: :cache, store: :store, raise_errors: true, type: nil
         define_method "remote_#{name}_url" do
         end
 
-        define_method "#{name}_url" do |*args|
-          Refile.attachment_url(self, name, *args)
+        define_method "#{name}_url" do |**args|
+          Refile.attachment_url(self, name, **args)
```
kahshing96, line 89: the `Refile.attachment_url` accept both `*args` and some other optional keyword arguments as well like `expires_at` etc
https://github.com/refile/refile/blob/c4ac577c6fdad92bc079a62a0e82888319daedc8/lib/refile.rb#L400

```suggestion
        define_method "#{name}_url" do |*args, **kwargs|
          Refile.attachment_url(self, name, *args, **kwargs)
```

vitalinfo: thanks, I've done single PR with all my latest update for Refile https://github.com/refile/refile/pull/612

vitalinfo:
move to https://github.com/refile/refile/pull/612

PR #612, opened 2024-01-27
Support Ruby 3+ and Rails 7+

vitalinfo:


Bramjetten:
Can this be merged? We'd like to migrate everything to ActiveStorage, but Refile is still keeping us on Ruby 2.7 in the meantime.

lib/refile/attachment/active_record.rb
```diff
@@ -17,8 +17,8 @@ def attachment(name, raise_errors: false, destroy: true, **options)
           if send(attacher).present?
             send(attacher).valid?
             errors = send(attacher).errors
-            errors.each do |error|
-              self.errors.add(name, *error)
+            errors.each do |error_type, error|
+              self.errors.add(name, error_type, **(error || {}))
             end
           end
```
leandro, line 23: **[Question/Suggestion]** Since I'm interested in your fix for this issue (which is happening on the application I work on as well -- Rails 6), I need to ask you this: are you sure this is going to work fine on Rails 7? I was checking its source code and the `ActiveModel::Errors#each` method implementation [is actually a method delegation](https://github.com/rails/rails/blob/v7.0.8.4/activemodel/lib/active_model/errors.rb#L78) to its `@errors` instance variable, which is a simple `Array` instance. So it looks like your changes wouldn't work well on Rails 7.

Nonetheless, it seems your implementation is actually based on Rails 5, where [ActiveModel::Errors#each](https://github.com/rails/rails/blob/v5.2.8.1/activemodel/lib/active_model/errors.rb#L167-L171) method implementation does expect a two parameters arity block, as you can see on the link I just shared. On the other hand, Rails 6's implementation of the same method will display a deprecation warning message when a block with two parameters arity is provided, as you can see [here in its source code](https://github.com/rails/rails/blob/v6.1.7.8/activemodel/lib/active_model/errors.rb#L238-L257).

So, having all that said, it's also worth to mention that we could go beyond the former version of this code and use [ActiveModel::Errors#import](https://api.rubyonrails.org/v6.1.7/classes/ActiveModel/Errors.html#method-i-import) method that is present on Rails 6+ (and Rails 7+) and use a fallback to the `add` method for older Rails versions.

Therefore, all things considereds, the change suggestion below should work at least on Rails 4+ to 7+:
```suggestion
          next if (attacher_record = send(attacher)).nil? || attacher_record.valid?

          # {ActiveModel::Errors#each} will yield only one parameter, a {ActiveModel::Error} object
          # as `attr_or_error` on Rails 6+ (and 7+), and two parameters on Rails < 6, with
          # `attr_or_error` being a {Symbol} object and `msg` being a {String} object.
          attacher_record.errors.each do |attr_or_error, msg|
            msg && errors.add(name, attr_or_error, message: msg) ||
              errors.import(attr_or_error, attribute: name)
          end
```

.gitignore
```diff
@@ -26,3 +26,5 @@ s3.yml
 spec/refile/test_app/log
 spec/refile/test_app/tmp
 spec/refile/test_app/db/*.sqlite
+
+.idea
```
leandro, line 30: Ooops!

leandro (changes requested):

leandro:
@jnicklas @sobrinho Is this project being actively maintained still? This PR is addressing a really important bug that I'd like to have fixed in the project I work on. Is there anything specific that we could do in order to move forward with this PR?

Bramjetten:
@leandro Fwiw, we decided to speed up the process of migrating to ActiveStorage. Using parallel jobs it was pretty doable with hundreds of thousands of uploads.

leandro:
> @leandro Fwiw, we decided to speed up the process of migrating to ActiveStorage. Using parallel jobs it was pretty doable with hundreds of thousands of uploads.

Gotcha! Thanks for the heads up, Bram. I might then open a new PR addressing that specific bug and an Issue associated to the new PR.

thadeu:
Hey folks, have you some ideia that when you send this to production?

thadeu:
Will resolve the problem that keyword: object? #618

Issue #618, opened 2025-01-14
 wrong number of arguments (given 3, expected 2; required keyword: object)


thadeu:
## Environment details

Ruby version: 3.2.3
Rails version: 7.1.5.1
Gem version: gem 'refile', github: 'refile/refile'

## Error details

Class: ActionView::Template::Error:
Message: wrong number of arguments (given 3, expected 2; required keyword: object)

Using SimpleForm simple like that:

```ruby
= f.input :options_csv, as: :attachment, direct: true
```

![image](https://github.com/user-attachments/assets/4cb6cafb-033e-4411-bc2c-249456c2936a)

If I use `= f.input :options_csv, object: { as: :attachment, direct: true }` will resolve my problem, ok!

## Question?

Why does I need to passa object keyword?

PR #615, opened 2024-07-01
Support Rails 6 and 7 on Ruby 3

leandro:
This fixes issue #614.

Issue #614, opened 2024-06-29
Bug: `ArgumentError: wrong number of arguments (given 3, expected 1..2)`


leandro:
## Environment details
Ruby version: `3.0.3`
Rails version: `6.1.7.7`
Refile version: `0.7.0` (rev: c4ac577)

## Error details
Class: `ArgumentError`
Message: `wrong number of arguments (given 3, expected 1..2)`
Stack trace:
```
activemodel (6.1.7.7) lib/active_model/errors.rb in add at line 404
refile (c4ac577c6fda) lib/refile/attachment/active_record.rb in block (2 levels) in attachment at line 21
refile (c4ac577c6fda) lib/refile/attachment/active_record.rb in each at line 20
refile (c4ac577c6fda) lib/refile/attachment/active_record.rb in block in attachment at line 20
```
I'm just pasting the 4 topmost lines, given most of the stack trace are all lines from other gems in general (Puma, NeweRelic, Rails and etc).

## Context
Basically, this error happens when there is any validation error on the attacher itself, which means validation errors on the Refile related attributes, such as `extension`, `content_type` and few others.

It doesn't look like it's a hard fix and I actually aim to open a PR for it, but it's basically what [I proposed in this review comment](https://github.com/refile/refile/pull/612#discussion_r1657656058) in PR #612.

1. **What did I do?**
1.1. A file with the wrong extension was provided to a model and I tried to save the model.
2. **What did you expect to happen?**
2.1. The validation would fail and a validation error message would be available through my model object.
3. **What happened instead?**
3.1. An application error (`ArgumentError` mentioned above) happened instead, breaking the submission request and returning an HTTP error response to end user instead.

thadeu:
+1 here

thadeu:
Reference issue: #618

PR #619, opened 2025-01-28
Compatibility to Rails 7.1.x and Ruby 3.3.x

thadeu:
# Why?

I think that repository was abandoned, the same way I'd created this PR to reforce our dependency to update this repository.

I need to skip some test in the features folder, because i dont know the test was broken.

# Related
#615 #612 #614 #618 #617 

# Tests

image

### DIFF
--- a/spec/refile/spec_helper.rb
+++ b/spec/refile/spec_helper.rb
@@ -4,6 +4,7 @@ require "refile"
 require "refile/backend_examples"
 require "webmock/rspec"
 require "refile/file_double"
+require 'byebug'
 
 tmp_path = Dir.mktmpdir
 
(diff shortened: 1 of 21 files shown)


spec/refile/test_app.rb
```diff
@@ -39,6 +42,16 @@ class TestApp < Rails::Application
   end
 end
 
+require 'capybara/poltergeist'
```
Faq, line 45: Actually poltergeist is long dead, what the point adding it here?

spec/refile/spec_helper.rb
```diff
@@ -4,6 +4,7 @@
 require "refile/backend_examples"
 require "webmock/rspec"
 require "refile/file_double"
+require 'byebug'
```
Faq, line 7: this should not be required

